PR #8, opened 2021-12-22
Accordion: Fix "t.map is not a function" error

emilhe:
In React, it is not safe to iterate the children attribute using map. Rather, the `React.Children.map` construct should be used,

https://stackoverflow.com/questions/34030328/how-to-iterate-through-a-components-children-in-typescript-react

Specifically, I get an error with the current code ("t.map is not a function") is a pass list with only one `AccordionItem` to the `Accordion` component. The PR fixes that error.


snehilvj:
Thanks a lot for the PR

PR #9, opened 2021-12-24
Add ScrollArea component

emilhe:
The PRs adds the `ScrollArea` component,

https://mantine.dev/core/scroll-area/

Here is a small example, which yields vertical and horizontal scroll bars,

```
from dash import Dash, html

content = """Charizard is a draconic, bipedal Pokémon. It is primarily orange with a cream underside from the chest 
to the tip of its tail. It has a long neck, small blue eyes, slightly raised nostrils, and two horn-like structures 
protruding from the back of its rectangular head. There are two fangs visible in the upper jaw when its mouth is 
closed. Two large wings with blue-green undersides sprout from its back, and a horn-like appendage juts out from the 
top of the third joint of each wing. A single wing-finger is visible through the center of each wing membrane. 
Charizard's arms are short and skinny compared to its robust belly, and each limb has three white claws. It has 
stocky legs with cream-colored soles on each of its plantigrade feet. The tip of its long, tapering tail burns with a 
sizable flame. """

# Small example app.
app = Dash()
app.layout = html.Div(
    dmc.ScrollArea(
        html.Div(content, style=dict(width="200px", height="200px")),
    ), style=dict(height="100px", width="100px"),
)

if __name__ == '__main__':
    app.run_server()
```

### DIFF
--- a/src/lib/components/ScrollArea.react.js
+++ b/src/lib/components/ScrollArea.react.js
@@ -9,7 +9,7 @@ import { pick } from "ramda";
 const ScrollArea = (props) => {
     return (
         <MantineScrollArea
-            {...pick(["dir", "offsetScrollbars", "scrollHideDelay", "scrollbarSize", "type"], props)}
+            {...omit(["setProps", "children"], props)}
         >
             {props.children}
         </MantineScrollArea>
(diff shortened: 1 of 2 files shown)


snehilvj (changes requested):
I think we should also pass style, id and className props to our MantineScrollArea component

emilhe:
Ah, yes. That was a mistake. It should be corrected now.

PR #12, opened 2021-12-28
Update ScrollArea.react.js

emilhe:
I just noticed that I forgot to update the import statement after the last update. It's corrected in this PR.


snehilvj:
I have corrected that in one of my branches, thanks.

PR #17, opened 2022-01-07
Update Chips component to use options/multi instead of data/multiple

shouples:
This is to keep keyword arguments more consistent with other packages' components like the [`dash_core_components`' Dropdown](https://dash.plotly.com/dash-core-components/dropdown) for `multi`, and the `dash_core_components` and `dash_bootstrap_components` Radio/Checklist/Switch for `options`.

Let me know if there is anything you would like changed here - it's my first contribution to a dash components package, and I would love to help get more of Mantine into `dash_mantine_components`! You've done a great job with the existing components so far. :clap:

Also, I was able to run a separate `usage.py` to test the `options/multi` changes:
```python
import dash_mantine_components
import dash
from dash import html, Input, Output


app = dash.Dash(__name__)

app.layout = html.Div([
    dash_mantine_components.Chips(
        options=[{'label': v, 'value': v} for v in list('abc')],
        multi=True,
        id='chips'
    ),
    html.Div(id='output')
])


@app.callback(
    Output('output', 'children'), 
    Input('chips', 'value')
)
def display_output(value):
    if not value:
        return 'Choose value(s)'
    return f'You have chosen {value}'


if __name__ == '__main__':
    app.run_server(debug=True)
```

But I was having selenium/chromedriver issues running `test_chips.py`.


emilhe:
I think the use of the keyword `data` is more in line with Mantine conventions, see e.g. the [select component](https://mantine.dev/core/select/). While I understand the desire to align with other Dash components, I generally prefer to keep naming conventions as close to the underlying React library as possible. That also makes it easier to reuse the React documentation.

snehilvj:
Thanks a lot @shouples for your interest in this library. However, in this case, I have to agree with @emilhe. In fact, in the beginning I had used `options` instead of `data` but as I added more components, I realised its better to keep dmc in sync with mantine.

shouples:
Thanks @emilhe and @snehilvj - makes sense! I'll keep that in mind for any future merge requests. 👍

PR #19, opened 2022-01-08
Add RingProgress component

emilhe:
This PR adds the [`RingProgress` component](https://mantine.dev/core/ring-progress/). Here is a small example,

```
from dash import Dash, html

app = Dash()
app.layout = html.Div([
    dmc.RingProgress(
        sections = [
            dict(value=50, color="red"),
            dict(value=50, color="blue"),
        ],
        label=[dmc.Center("Label", style=dict(width="100%"))]
    ),
])

if __name__ == '__main__':
    app.run_server(port=7777)
```

image


emilhe:
I was actually a little in doubt about the `color` property. As I read the docs, the available colors depends on the theme. So maybe the `PropType` should just be `Proptypes.string`, and there should be some comment that elaborates a bit (e.g. with the colors available in the default theme, a link, or so). What do you think @snehilvj ?

snehilvj:
Two things here:
1. color prop: A lot of other components only support colors from Mantine theme. This is one of the exceptions. I think, specifying it like this would form the correct docstring:
```javascript
color: PropTypes.oneOfType([
        PropTypes.string,
        PropTypes.oneOf([
            "dark",
            "gray",
            "red",
            "pink",
            "grape",
            "violet",
            "indigo",
            "blue",
            "cyan",
            "teal",
            "green",
            "lime",
            "yellow",
            "orange",
        ]),
    ]),
```

2. Can you please help me understand the use of renderDashComponent from dash-extensions-js?

snehilvj:
Also, going forward I'd like to follow some kind of system for new contributions (maybe like dev-master) instead of merging all changes directly to master. Any inputs, you'd like to provide?

emilhe:
1) Ah, interesting. I didn't notice that. In that case, I guess your proposed solution is better. However, I can't keep thinking that it feels not-so-nice to keep repeating the list of theme colors. Couldn't we figure out something better? 

One option could be just to use `PropTypes.string` and then create an `Enum` (in Python) with the color values supported as part of the dmc library. It could be named `Color` (or maybe `ThemeColor`), e.g. something like 

```
class Color(Enum):
    RED = "red"
    BLUE = "blue"
    ...
```

in the library, so that the syntax would be along the lines of,

```
from dash import Dash, html

app = Dash()
app.layout = html.Div([
    dmc.RingProgress(
        sections = [
            dict(value=50, color=dmc.Color.RED),
            dict(value=50, color=dmc.Color.BLUE),
        ],
        label=[dmc.Center("Label", style=dict(width="100%"))]
    ),
])

if __name__ == '__main__':
    app.run_server(port=7777)
```

What do you think?

2) Per default, Dash only renders Dash objects passed via the `children` property. Hence passing a Dash object via a non-children property like `label`,

`label=[dmc.Center("Label", style=dict(width="100%"))]`

will not work per default. The `renderDashComponent` function turns the json structure received in the js layer into the desired React node, similar to the internals of the dash renderer itself, thus allowing (with a few line of code) to omit this limitation.

emilhe:
@snehilvj This is also a discussion we are currently having at my workplace. At the moment, we are leaning towards **not** using a development branch, but rather only merge in feature branches when they are mature enough to be released. I am interested in hearing your thoughs as well.

snehilvj:
> will not work per default. The `renderDashComponent` function turns the json structure received in the js layer into the desired React node, similar to the internals of the dash renderer itself, thus allowing (with a few lines of code) to overcome this limitation. The limitation of this approach that i know of is that you can't attach callbacks to Dash objects that are rendered this way.

Wow 🤯 this exists.. I need to go through all the existing components and refactor them.

On the issue of colors, we need to think more about it. The solutions needs to account for the entire theme not just colors, see: https://mantine.dev/theming/mantine-provider/
Here you can also see that this theme can be customised. In that case, hardcoded enum might not be the way to go.
The solution needs to be more dynamic. I'd think more about this and might create a new issue on this.

Regarding the discussion on development branch, keeping master stable is not a requirement here as such, but I think it will be easier to track the changes between different releases.

emilhe:
It exists! With emphasis on presence ;). I made it recently, so it probably didn't exist when you made the initial implementation. But yes, I agree that it would be really nice to refactor a number of components to use the strategy. It will improve flexibility and keep the dmc better in sync with mantine.

About tracking changes, I have been really happy about simply creating a `CHANGELOG.md` file, e.g. this one

https://github.com/thedirtyfew/dash-leaflet/blob/master/CHANGELOG.md

A pull request that e.g. adds a component (like this one) must then add an appropriate statement to this file prior to merging. That's also how it works making pull requests to the main Dash repo. If you are thinking more along the lines of "code changes", you can simply do a compare between tags on the master branch,

https://github.com/snehilvj/dash-mantine-components/compare/0.2.1...0.2.2

snehilvj:
yeah I think this will be fine. I'll start a changelog starting next release.

snehilvj:
I think renderDashComponents needs more work at this stage. I frequently get this error: 
`Warning: li: 'key' is not a prop. Trying to access it will result in 'undefined' being returned. If you need to access the same value within the child component, you should pass it as a different prop.`

Lets hold adding dash-extensions-js to dmc for now. I'll also create an issue in the dash-extensions-js repo with some sample code.

PR #22, opened 2022-01-09
Add JsonInput component

emilhe:
This PR adds the [`JsonInput` component](https://mantine.dev/core/json-input/). Here is a small example,

```
import json
from dash import Dash, html

data = dict(fruits=["apple", "orange", "banana"], vegetables=["tomato", "onion", "lettuce"])

app = Dash()
app.layout = html.Div([
    dmc.JsonInput(value=json.dumps(data),
                  label="Some data",
                  validationError="Invalid JSON",
                  required=True)
])

if __name__ == '__main__':
    app.run_server(port=7777)
```

image


src/lib/components/JsonInput.react.js
```diff
@@ -0,0 +1,179 @@
...
+/**
+ * JsonInput. For more information, see: https://mantine.dev/core/json-input/
+ */
+const JsonInput = (props) => {
+    const { class_name, children, setProps } = props;
+    let nProps = omit(["setProps", "children", "class_name"], props);
+    // Render react nodes.
+    nProps = renderDashComponents(nProps, ["icon", "rightSection", "validationError", "description", "error", "label"]);
+    // Bind OnChange event.
+    nProps.onChange = value => setProps({ value });
```
snehilvj, line 16: Don't think this will work.

emilhe: It does :)

PR #30, opened 2022-01-28
Add NumberInput.react.js

emilhe:
This PR adds the [`NumberInput` component](https://mantine.dev/core/number-input/). Here is a small example,

```
from dash import Dash, html

app = Dash()
app.layout = html.Div([
    dmc.NumberInput(value=5, min=0, max=10)
])

if __name__ == '__main__':
    app.run_server(port=7777)
```

image


snehilvj:
Thanks, this will be part of the next release 0.4.0.

PR #32, opened 2022-01-29
Add time input component

emilhe:
This PR adds the [`TimeInput` component](https://mantine.dev/dates/time-input/). Here is a small example,

```
from dash import Dash, html, Input, Output
from datetime import datetime

app = Dash()
app.layout = html.Div([
    dmc.TimeInput(value=datetime.now().isoformat(), id="time", withSeconds=True),
    html.Div(id="log")
])


@app.callback(Output("log", "children"), Input("time", "value"))
def update(time):
    return datetime.fromisoformat(time)


if __name__ == '__main__':
    app.run_server(port=7777, debug=True)
```

image

### DIFF
--- a/src/lib/components/TimeInput.react.js
+++ b/src/lib/components/TimeInput.react.js
@@ -1,0 +1,210 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { omit } from "ramda";
+import { TimeInput as MantineTimenInput } from "@mantine/core";
+import {renderDashComponents} from "dash-extensions-js"
+
+/**
+ * JsonInput. For more information, see: https://mantine.dev/core/json-input/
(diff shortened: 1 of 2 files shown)


emilhe:
I guess this component is the one that's least intuitive to use of my recent PRs. To adhere to the mantine convention of having time represented via a `Date ` object, the user must pass an iso formatted datetime string as value (not just the time), similarly a datetime iso string is returned. It works (and it is coherent with mantine), but I am very open for suggestions on a better approach.

snehilvj:
Let me merge it for now, and I'll play around with the component. Will let you know my feedback.

snehilvj:
No feedback as such. The component works fine.

PR #36, opened 2022-02-02
Add Mantine default colors

jlfsjunior:
This is a tiny PR to add the [default colors](https://mantine.dev/theming/extend-theme/#default-colors) as a Python object. It can be quite useful when composing themes.

I am not entirely convinced that a dict is the best representation, it could be nicer to access each list per attribute (`colors.red` vs `colors["red"]`). Any thoughts are appreciated!

As this is the first non-component submodule, I am also 100% open to a better structure. To me it makes sense to add to `theme/__init__.py`, so it can be imported using `dash_mantine_components.theme`. I am happy to change otherwise.


snehilvj:
Rearding colors.red vs colors["red"], I think the latter is good enough.
With the first structure:
1. we need to write extra code to make it an attribute dict
2. for colors like "sky-blue", "french-plum", etc. , we need to access them using the dictionary style anyway.

jlfsjunior:
Thanks for the input, @snehilvj!

I haven't thought about two-word names for colors, it is a good call.

I will make the changes and let you know. Thanks again!

snehilvj:
Hey @jlfsjunior, were you able to make any progress on this, I might do a release this weekend, so would be good to include this as well. If you want, I can make the remaining changes myself.

jlfsjunior:
Thanks for the reminder @snehilvj. I made a slight modification in the import, as the relative one wasn't working.

snehilvj:
This will be released in 0.4.1

PR #41, opened 2022-02-22
Popover component

kaburelabs:
This PR is about the Popover component;

It's not 100% ready because it has a property called TARGET which receives a component;

image

I didn't understand how can I insert the "onClick" function inside the received component... If someone can provide some support on this I would appreciate a lot


kaburelabs:
If you want to test the Popover in Dash you can use the following code:

```
import dash

external_stylesheets = ["https://unpkg.com/tailwindcss@^2/dist/tailwind.min.css"]
app = dash.Dash(__name__, external_stylesheets=external_stylesheets)

app.layout = html.Div(
    className="text-center",
    children=[
        html.Div("PopOver test"),
        dmc.Popover(
            children=[html.Div("testing Other thicrewa", style={"background": "red"})],
            opened=False,
            radius="lg",
            placement="start",
            width=300,
            spacing="lg",
            title="BLABALAO FKODS",
            id="popover-test",
        ),
        html.Button(id="my-custom-btn", children=["name test"]),
        html.Div(id="my-custom-component"),
    ],
    style={"padding": "64px"},
)

if __name__ == "__main__":
```

PR #42, opened 2022-02-23
Popover Component 

kaburelabs:
This PR is to add the Popover Component but the target parameter is not correctly configured yet... It's still hardcoded with a button as in the Popover example;   

image

I would appreciate a lot any collaboration to correctly set this


kaburelabs:
To test the component you can use the following snippet.

```
import dash

app = dash.Dash(__name__)

app.layout = html.Div(
    className="text-center",
    children=[
        html.Div([
            html.Div("PopOver test"),
            dmc.Popover(
                children=[html.Div("Testing component Children", 
                                   style={"background": "red"})],
                opened=False,
                radius="lg",
                # target=[dmc.Button(id="tests", children="children1")],
                placement="start",
                width=300,
                spacing="lg",
                title="Popover Title Test",
            ),

            html.Div(id="my-custom-component")
        ], style={"marginBottom":"64px"})],
    style={"padding": "64px"},
)


if __name__ == "__main__":
```

snehilvj:
Thanks @kaburelabs but I'll add it in one of the next releases. Can't use this right now as I have migrated to Typescript and Mantine v5.

PR #43, opened 2022-02-24
Adding Menu, MenuItem and MenuLabel

kaburelabs:
In the Menu Component, the unique problem I found is that the MenuItem seems like it's not rendering the icon... I don't know if I did something wrong but would be great if someone can validate it ;
<br>

To test the component you can use the reproducible example below:

```
import dash

app = dash.Dash(__name__)

app.layout = html.Div(
    className="text-center",
    children=[
        html.Div([
            html.Div("Menu and sub Menu Components test"),
            dmc.Menu(children=[dmc.MenuLabel(children="Label test"),
                               html.Div(children=['"Div Option"']),
                               dmc.MenuItem(children="MenuItem1"),
                               dmc.Divider(),
                               dmc.MenuItem(children="testing"),
                               dmc.MenuItem(children=["Testing with div"])]),
        ]),
    ],
    style={"padding": "64px"},
)

if __name__ == "__main__":
```


snehilvj:
What about the callbacks? Can you please also include how this component will be used in the apps?

kaburelabs:
Hey @snehilvj I did some modifications and now it's a good alternative to create links to redirect the user as a button; 

image

PR #52, opened 2022-02-27
Highlight component first implementation

kaburelabs:
How can I add the prop Component? 

image

To test the component you can use the below snippet:

```

import dash
from dash import html

app = dash.Dash(__name__)

app.layout = html.Div(
    className="text-center",
    children=[
        html.Div([
            html.Div("Highlight Component test"),
            dmc.Highlight(children="Highlight This, definitely THIS and also this!", 
                          highlight="this", 
                          align="right", size="xl",
                        #   transform="uppercase", 
                        #   underline=True, 
                          inherit=False,
                          color="red",
                          inline=True,
                          highlightColor="green")
        ]),
    ],
    style={"padding": "64px"},
)

if __name__ == "__main__":
```

### DIFF
--- a/src/lib/components/Highlight.react.js
+++ b/src/lib/components/Highlight.react.js
@@ -1,0 +1,148 @@
+import React from "react";
+import { Highlight as MantineHighlight } from '@mantine/core';
+import PropTypes from "prop-types";
+import { omit } from "ramda";
+
+/**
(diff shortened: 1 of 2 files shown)


snehilvj:
Let's leave this prop for on of the future upgrades.

PR #54, opened 2022-03-04
fixing the bug related to the issue 33 & 47

kaburelabs:
Related to the datepicker opened issues


snehilvj:
Thanks @kaburelabs for the PR. But this is not really the right solution. This change will prevent users from settings dates via callbacks.
However, thanks to your logs on our discord channel, I think I have a solution. I have pushed the changes to the master, can you please pull the latest and test it out?

PR #55, opened 2022-03-05
Adding component Mark

kaburelabs:
It's a simple component that highlights a text

image

```
from datetime import datetime, date

from dash import Input, Output, html
import dash

app = dash.Dash(__name__)

app.layout=html.Div(
    [
        # dmc.TransferList(value=[[],[]]),
        html.Div(["Hellow, testing ", dmc.Mark(" Mark component ", color="grape"), " as a test." ]),
    ]
)

if __name__ == "__main__":
```

### DIFF
--- a/src/lib/components/Mark.react.js
+++ b/src/lib/components/Mark.react.js
@@ -1,0 +1,63 @@
+import React, { useState } from "react";
+import { Mark as MantineMark } from '@mantine/core';
+import PropTypes from "prop-types";
+import { omit } from "ramda";
+
+
+/**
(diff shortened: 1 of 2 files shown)


src/lib/components/Mark.react.js
```diff
@@ -0,0 +1,63 @@
...
+        "violet",
+        "indigo",
+        "blue",
+        "cyan",
+        "teal",
+        "green",
+        "lime",
+        "yellow",
+        "orange",
+    ]),
```
snehilvj, line 54: Missing props: `class_name`, `style`.

PR #56, opened 2022-03-05
changing children from string to node to remove a warning it was raising

kaburelabs:
I've just changed the children type to node

image


snehilvj:
Corrected this for the future release.

PR #60, opened 2022-03-17
Fixes bug; chip disabled state not passed to component

emilhe:
Currently, it is not possible to set the `disabled` state for chips, as the state is not passed to the react layer. This PR fixes the bug.

```
from dash import Dash

app = Dash()
app.layout = dmc.Chips(
    data=[{"value": c, "label": c, "disabled": c == "£"} for c in ["€", "£", "$", "¥"]]
)

app.run_server(port=7779)
```

### DIFF
--- a/src/lib/components/Chips.react.js
+++ b/src/lib/components/Chips.react.js
@@ -21,7 +21,7 @@ const Chips = (props) => {
         >
             {data.map((chip, index) => {
                 return (
-                    <Chip value={chip.value} key={index}>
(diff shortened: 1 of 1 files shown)


Roh-codeur:
@snehilvj : hey buddy, how is it going? I need this as well. Do you think its a good time to roll this out please? 

thanks for all your help with this lib!

emilhe:
I believe this change will be included in the next release, which is expected around 27/3 :)

Roh-codeur:
Awesome, thanks! Look forward to it

PR #83, opened 2022-07-08
Add the control property thanks to component as props

RenaudLN:
Adding a `control` property with the new component property for the Menu allowing to get a menu with different target than the default '...'

image

For discussion: I tried without the wrapping div but then the click event would not get passed on to the Dash provided control element.


snehilvj:
Thanks a lot @RenaudLN but I have added this in dmc (which is in typescript and is based on Mantine 5).

PR #84, opened 2022-07-14
Multi-select working with creatable option

RenaudLN:
dmc_multiselect_create


snehilvj:
Thanks @RenaudLN. I'll make sure its available in the next dmc release.

PR #107, opened 2022-11-09
Add Card component; CardSection is work-in-progress

nlyle01:
Submitting this as a **draft** for discussion. Hoping to get some input on how to implement `CardSection` (otherwise `Card` by itself is just `Paper` with a border).

Per Mantine's docs, `CardSection` [is a special component that is used to remove Card padding from its children while other elements still have horizontal spacing](https://mantine.dev/core/card/#cardsection). The screenshot below is my `usage.py` app; You can see how the image in the first card, and the header in the second card, correctly fill the horizontal space but not the vertical space to the tops of the cards. The good news is that the other Card Section props like `inheritPadding` and `withBorder` work as expected. 

Screen Shot 2022-11-08 at 4 05 39 PM

Here's my `usage.py` to produce the cards in the screenshot:

```python
from dash import Dash, html

app = Dash(__name__)

card_example = dmc.Card(
    children=[
        dmc.CardSection(
            dmc.Image(
                src="https://images.unsplash.com/photo-1527004013197-933c4bb611b3?ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&ixlib=rb-1.2.1&auto=format&fit=crop&w=720&q=80",
                height=160,
            ),
        ),
        dmc.Group(
            [
                dmc.Text("Norway Fjord Adventures", weight=500),
                dmc.Badge("On Sale", color="red", variant="light"),
            ],
            position="apart",
            noWrap=True,
            mt="md",
            mb="xs",
        ),
        dmc.Text(
            "With Fjord Tours you can explore more of the magical fjord landscapes with tours and activities on and around the fjords of Norway",
            size="sm",
            color="dimmed"
        ),
        dmc.Button(
            "Book classic tour now",
            variant="light",
            color="blue",
            fullWidth=True,
            mt="md",
            radius="md",
        ),
    ],
    withBorder=True,
    shadow="sm",
    radius="md",
)

images = [
  dmc.Image(radius="sm", src='https://images.unsplash.com/photo-1449824913935-59a10b8d2000?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=250&q=80'),
  dmc.Image(radius="sm", src='https://images.unsplash.com/photo-1444723121867-7a241cacace9?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=250&q=80'),
  dmc.Image(radius="sm", src='https://images.unsplash.com/photo-1444084316824-dc26d6657664?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=250&q=80'),
]

card_example_2 = dmc.Card(
    children=[
        dmc.CardSection(
            dmc.Group(
                children=[
                    dmc.Text("Review Pictures", weight=500),
                    dmc.ActionIcon(DashIconify(icon="carbon:overflow-menu-horizontal"), color="gray", variant="transparent"),
                ],
                position="apart"
            ),
            withBorder=True,
            inheritPadding=True,
            py="xs"
        ),
        dmc.Text(
            children=[
                dmc.Text("200+ images uploaded since last visit, review them to select which one should be added to your gallery"),
            ],
            mt="sm",
            color="dimmed",
            size="sm"
        ),
        dmc.CardSection(
            dmc.Image(
                src="https://images.unsplash.com/photo-1579263477001-7a703f1974e6?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=500&q=80",
                mt="sm",
            ),
        ),
        dmc.CardSection(
            children=[
                dmc.SimpleGrid(
                    cols=3,
                    children=[i for i in images]
                ),
            ],
            inheritPadding=True,
            mt="sm",
            pb="md"
        ),
    ],
    withBorder=True,
    shadow="sm",
    radius="md",
)

app_layout = html.Div(
    [
        html.Br(),
        html.Br(),
        dmc.Container(
            children=[
                dmc.Grid(
                    children=[
                        dmc.Col(card_example, span=6),
                        dmc.Col(card_example_2, span=6),
                    ],
                    gutter="xl"
                ),
            ]
        )
    ]
)

    children=app_layout,
    withNormalizeCSS=True,
    withGlobalStyles=True,
    # theme={"colorScheme": "dark"},
)

if __name__ == "__main__":
    app.run_server(debug=True)

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,11 @@ This is an alpha release based on Mantine v5. This version of dmc is a big overh
 
 - Tooltip component was not working properly.
 
+### Added
+
+- New components:
+  - Card (and CardSection)
+
 # 0.10.1
 
 ### Fixed
--- a/src/ts/components/core/card/Card.tsx
+++ b/src/ts/components/core/card/Card.tsx
@@ -1,0 +1,30 @@
+import React from "react";
+import { DefaultProps } from "../../../props";
+import { Card as MantineCard } from "@mantine/core";
+import { MantineNumberSize, MantineShadow } from "@mantine/styles";
+
+type Props = {
+    /** Card children */
+    children?: React.ReactNode;
+    /** Predefined border-radius value from theme.radius or number for border-radius in px */
+    radius?: MantineNumberSize;
+    /** Predefined box-shadow from theme.shadows (xs, sm, md, lg, xl) or any valid css box-shadow property */
+    shadow?: MantineShadow;
+    /** Adds 1px border with theme.colors.gray[3] color in light color scheme and theme.colors.dark[4] in dark color scheme */
+    withBorder?: boolean;
+} & DefaultProps;
+
+/**
+ * Renders white or dark background depending on color scheme. For more information, see: https://mantine.dev/core/card/
+ */
+ const Card = (props: Props) => {
+    const { children, setProps, ...other } = props;
+
+    return <MantineCard {...other}>{children}</MantineCard>;
+};
+
+Card.defaultProps = {
+    p: 'md',
+};
+
+export default Card;
--- a/src/ts/components/core/card/CardSection.tsx
+++ b/src/ts/components/core/card/CardSection.tsx
@@ -1,0 +1,28 @@
+import React from "react";
+import { DefaultProps } from "../../../props";
+import { CardSection as MantineCardSection } from "@mantine/core";
+
+type Props = {
+    /** Card children */
+    children?: React.ReactNode;
+    /** Determines whether section from inherit padding from Card */
+    inheritPadding?: boolean;
+    /** Adds 1px border with theme.colors.gray[3] color in light color scheme and theme.colors.dark[4] in dark color scheme */
+    withBorder?: boolean;
+} & DefaultProps;
+
+/**
+ * Renders white or dark background depending on color scheme. For more information, see: https://mantine.dev/core/card/
+ */
+ const CardSection = (props: Props) => {
+    const { children, setProps, ...other } = props;
+
+    return <MantineCardSection {...other}>{children}</MantineCardSection>;
+};
+
+CardSection.defaultProps = {
+    withBorder: false,
+    inheritPadding: false,
+};
+
+export default CardSection;
--- a/src/ts/index.ts
+++ b/src/ts/index.ts
@@ -15,6 +15,8 @@ import Blockquote from "./components/core/Blockquote";
 import Breadcrumbs from "./components/core/Breadcrumbs";
 import Button from "./components/core/button/Button";
 import ButtonGroup from "./components/core/button/ButtonGroup";
+import Card from "./components/core/card/Card";
+import CardSection from "./components/core/card/CardSection";
 import Center from "./components/core/Center";
 import Checkbox from "./components/core/checkbox/Checkbox";
 import CheckboxGroup from "./components/core/checkbox/CheckboxGroup";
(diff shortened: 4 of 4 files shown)


mikejturner:
I looked into the issue with the image not filling all the vertical space at the top of the card.

This is how it works in pure react app (i.e. one not using dash):

The `Card` component looks at it's children. If the first child is a `CardSection` component then a data attribute `data-first` gets added to the `CardSection`. Similarly the `data-last` data attribute is applied to the last child if it is a `CardSection` component. (See [code block](https://github.com/mantinedev/mantine/blob/0e31440fbcd24256cc89fad32043df3a3df310e2/src/mantine-core/src/Card/Card.tsx#L29 )). These [data attributes apply some CSS](https://github.com/mantinedev/mantine/blob/0e31440fbcd24256cc89fad32043df3a3df310e2/src/mantine-core/src/Card/CardSection/CardSection.styles.ts#L28) that, for example, applies a negative margin which removes that empty space at the top of the card. It is important to realize that these data attributes will only get applied to `CardSection` components and not any other components that happen to be in the first and last child positions.

We know the above isn't working with your Dash components. But why? The reason is that Dash is wrapping extra React components around the `CardSection` component. When the card looks at it's first child, it doesn't see the CardSection component but a component introduced by dash when wrapping the CardSection. Since the first child is no longer a CardSection, the data attributes don't get applied and by extension the extra CSS styles to remove the gap at the top of the card.

We can use the React developer tools (in firefox) to compare at the component hierarchies for both a pure react app and your dash example. 

Here's the component hierarchy for the pure react example taken from card example (https://mantine.dev/core/card/):

ReactToolsMantineComponentHierarchy

We can see the `CardSection` as the first child of the `Card` (since the context provider, @mantine/core/paper and @mantine/core/box are the internals of how `Card` renders itself and children passed to it)

If we look at component hierarchy for the card you created in your usage.py we see something different:

ReactDevToolsDashUsageComponentHierarchy

There is the `Card` and `CardSection` but the `CardSection` is no longer the first child. Instead it is wrapped in 5 extra components (`Anonymous`, `Context.Consumer`, `BaseTreeContainer`, `ComponentErrorBoundary` and `CheckComponent`). It's these wrapping components that break the functionality in your Card/CardSection components.

Is there a fix?
To be honest I've really struggled to come up with one. Realistically I don't think we can change [Mantine Components](https://github.com/mantinedev/mantine) to somehow recognize dash-wrapped components and do the right thing.

I thought that maybe we add the data attributes (`data-first` and `data-last`) manually ourselves. But I don't think that will work because the mantine CardSection component doesn't have a props for them and again we don't really want to be changing Mantine Components itself.

Another option (but this is getting really hacky) is a add manually add a class to the CardSection say `className="data-first"` and then use some javascript to apply the `data-first` data attribute after the page has loaded. I haven't tried this though.

I'm open to suggestions/ideas.

nlyle01:
Thanks @mikejturner, really interesting to see the comparison between component hierarchies. I also thought about adding some of our own logic to apply negative margins id `Card` child is first or last element, but wasn't sure about the best approach.

Perhaps at this time, we use `Paper` with a border to mimic the look of a card, and if we want the `CardSection` effect of ignoring the `Card` padding, we nest `Paper` where the inner `Paper` has padding. I do this in one of my apps and it works okay.

@snehilvj curious what you think once you have a chance to play with it.

snehilvj:
This is not the first time I have encountered this situation. The fix is somewhat simple if you have done this before.

Issue #34, opened 2022-02-02
[Feature Request] Allow Multiselect to trigger callbacks on dropdown close.


snehilvj:
As suggested in [forum](https://community.plotly.com/t/multi-select-dropdown-that-doesnt-run-callback-until-all-choices-selected/60613),
Adding a `dropdownClosed` prop will enable triggering callbacks after a user is done selecting all the desired values.

Lxstr:
@snehilvj Hi there. I'd be interested to help implement this!

AnnMarieW:
Rather than adding another prop like `dropdownCose` it would be better to handle this with `debounce=True`

Check out this example based on  PR #383

https://py.cafe/amward/dmc-pr-383-demo-multi-select

Lxstr:
Amazing work!

PR #136, opened 2023-01-11
Added `isOpen` state to trigger dropdown close using `onclose` attribute 

D3nii:
isOpen state is a state which is true if the state of the dropdown is active, else it is false. This will be used to handle callback functions more effectively incase user wants to select multiple inputs, as the normal callback of `value` will be triggered on each select/deselect, `isOpen` can be used to minimize excess computation.

Corresponds to https://github.com/snehilvj/dash-mantine-components/issues/34

### DIFF
--- a/src/ts/components/core/select/MultiSelect.tsx
+++ b/src/ts/components/core/select/MultiSelect.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import {
     DefaultProps,
     PersistenceProps,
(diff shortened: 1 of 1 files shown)


Lxstr:
Would it also make sense to also (or instead) have a prop when can add which simply only changes the value state when the dropdown is closed? Ie:

```
dmc.MultiSelect(updateOnClose=True)
```

yeban:
This looks promising. How to use it? I tried the following, but `isOpen` is always set to `undefined` in the callback.
```
@callback(
Output(...)
Input('multiselectId', 'value')
State('multiselectId', 'isOpen')
)
def callback(value, isOpen):
  ...
```

Lxstr:
This wasn't merged in the end, I'm guessing due to work on v6 dmc.
Can I suggest trying out my version at [https://github.com/Lxstr/dash-mantine-components/tree/multiSelect-update-on-close](https://github.com/Lxstr/dash-mantine-components/tree/multiSelect-update-on-close) (build and install locally) and see if that works for you?

yeban:
Ah, silly me. Thanks.

Issue #140, opened 2023-01-17
CheckboxGroup default selection doesn't work


emilhe:
I was trying to create a `CheckboxGroup `with one of the checkboxes ticked by default. It was unclear to me, if I should use the checked property of the `Checkbox `, or the value property of the `CheckboxGroup`. However, non of them seems to work. Here is a MWE demonstrating both approaches,

```
from dash import Dash
import dash_mantine_components as dmc

app = Dash(__name__, )
app.layout = dmc.CheckboxGroup([
    dmc.Checkbox(value="x", label="X", checked=True),
    dmc.Checkbox(value="y", label="Y", checked=False),
], value=["x"])

if __name__ == "__main__":
    app.run_server()
```
![image](https://user-images.githubusercontent.com/1623542/212871925-ef43b9e7-7979-4f1c-9bda-a04909d25d07.png)

snehilvj:
Thanks @emilhe for reporting this. Its fixed and will be available in the next release.

PR #141, opened 2023-01-17
Fix CheckboxGroup default selection

emilhe:
This PR fixes the bug [#140](https://github.com/snehilvj/dash-mantine-components/issues/140). After applying the fix, the default selections are now reflected correctly.

image


snehilvj:
The problem was that value prop was not passed to the Checkbox group component. Its a controlled component, so we will use the `value` prop and not `defaultValue` prop.

Issue #173, opened 2023-04-11
Add n_submits property to TextInput


jacobswe:
The [dcc.Input](https://dash.plotly.com/dash-core-components/input) component has an n_submits property that increments when the <kbd>Enter</kbd> key is pressed while the input has focus. This would be useful for example in the TextInput or PasswordInput to quickly capture the values without requiring a separate button push (e.g., messaging). I'm happy to give this a shot if you can point me in the right direction.

AnnMarieW:
Hi @jacobswe 

Thanks for raising this issue and doing the PR.  Sorry for the long delay.

We are looking into developing a comprehensive and consistent  way to add debounce to all the input and select components like in the dash-core-components and dash-boostrap-components libraries.  I'm going to close this and the related PR for now and so it can be combined with #87. 

Please feel free to chime in on that issue and if you are still interested in doing a PR and once we have a plan, I'd be delighted.

Thanks again for your contribution.,

Issue #87, opened 2022-08-11
[Feature Request] Input/Select components add debounce=True,  n_submit and n_blur prop to trigger callback 


hrithikgavali:
Hi there,

I switched my app to use Dash Mantine Components because of the beautiful components UI and flexibility. However, I miss the ability to debounce callbacks from input and dropdowns in dash core components. Debounce allows us to trigger callbacks on the input after losing focus or pressing Enter Key, instead of firing the callback for each letter typed in the input box.

1. Is there a way to implement this with the current codebase?
2. If not, is there a way to style a component from dash core components package, exactly like dash mantine components, using some kind of a wrapper?

mcpcpc:
I second the need for this! this would definitely make large pattern-matching callbacks “feel” more performant.

snehilvj:
I added ability to debounce callbacks by a certain duration, i.e., you can specify `debounce=600` in your component and it will debounce callbacks by 600ms. Does it solve your purpose?

knightjdr:
The debounce you've added is a useful feature, and technically what debounce should be doing. However, the debounce property in Dash Core Components will trigger a callback only when focus is lost or the enter key is pressed, rather than being based on a time interval. They shouldn't have called it debounce, but nevertheless.

The feature you have added will work for me for the time being, but if you are able in the future to add the Dash version of debounce, that would be greatly appreciated. Maybe just call it something else.

BSd3v:
Currently, I cannot use this version of debounce because it triggers the callback and I have a running state that causes the input to disappear from the field of view.

emillma:
I agree that it would be nice if dmc gets the same debounce feature as dcc and dbc. I miss that pressing enter triggers the callback. 
As a quick fix; you can get the same debounce behavior using [EventListener](https://www.dash-extensions.com/components/event_listener) and listen for the [change](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event) event.

sergei3000:
+1 for the "on change" kind of behavior. This would be much appreciated.

jbeyer16:
I too am in support of enabling debounce to work in the same fashion as DCC and DBC for a consistent interface across components. 

In the meantime - it appears to me that, while `debounce=1000` works for `dmc.TextInput` it does not appear to work for `dmc.TimeInput` in that the callback is immediately fired upon edit. This makes it hard to edit the time values as desired.

etiennecelery:
[...]

PR #177, opened 2023-04-12
Implements Issue #173: Add n_submits property to TextInput

jacobswe:
This implements the issue I raised on the Discord and then in #173. Essentially, the goal is to be able to replicate the behavior of [dcc.Input](https://dash.plotly.com/dash-core-components/input), specifically how it handles what it calls `n_submit`, which looking through their source code is essentially just incrementing a count of the number of times the enter key has been pressed.

I've gotten this working locally on my end. It seems like a pretty trivial change... but these are also the first lines of TypeScript that I've ever written, so feel free to let me know if there are any issues here. Really love the package and use it for everything, so I'm happy to have gotten a reason to dig a little deeper into how it all works.

MRE:
```
from dash import Dash, html, callback, Output, Input, State

app = Dash(__name__)

app.layout = html.Div([
    dmc.TextInput(id='text-input'),
    dmc.Text('Nothing yet submitted', id='text-output')
])

@callback(
    Output('text-output', 'children'),
    Input('text-input', 'n_submit'),
    State('text-input', 'value'),
    prevent_initial_call=True
)
def test_textinput_n_submit(n_submit, text_input):
    return text_input

if __name__ == '__main__':
    app.run_server(debug=True)
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "dash_mantine_components",
-    "version": "0.12.0",
+    "version": "0.12.1",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "dash_mantine_components",
-            "version": "0.12.0",
+            "version": "0.12.1",
             "license": "MIT",
             "dependencies": {
                 "@emotion/react": "^11.10.4",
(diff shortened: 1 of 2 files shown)


AnnMarieW:
Just repeating the comment that I posted in the related issue 

>Thanks for raising this issue and doing the PR. Sorry for the long delay.
>
> We are looking into developing a comprehensive and consistent way to add debounce to all the input and select components like in the dash-core-components and dash-boostrap-components libraries. I'm going to close this and the related PR for now and so it can be combined with https://github.com/snehilvj/dash-mantine-components/issues/87.
>
>Please feel free to chime in on that issue and if you are still interested in doing a PR and once we have a plan, I'd be delighted.

Thanks again for your contribution.,

PR #202, opened 2023-08-18
Show Menu.Item as disabled when argument is set to True

RenaudLN:


snehilvj:
Thanks @RenaudLN for this change.

PR #270, opened 2024-05-21
Add prop setter for stepper active state

ondave:
Any reason not to add a prop setter for the active state?. Then you can do something like this in response to interacting with the stepper directly:
```

@callback(
    Output("my-text-element", "value"),
    Input("stepper", "active"),
    prevent_initial_call=True,
)
def update_stepper(current_step):
    return f"I am on step {current_step}"
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "dash_mantine_components",
-    "version": "0.14.1",
+    "version": "0.14.3",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
             "name": "dash_mantine_components",
-            "version": "0.14.1",
+            "version": "0.14.3",
             "license": "MIT",
             "dependencies": {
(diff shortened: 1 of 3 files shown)


ondave:
BTW - thanks for this library - I am a big fan of mantine, great to have for Dash as well and up to date with v7.

snehilvj:
Thanks @ondave for your feedback and this PR. I'm happy to merge this PR if you can update it to only include the .tsx file.

AnnMarieW:
@ondave 

Thanks for this PR! I was reviewing it to assist  getting it across the finish line.   However, it looks like you can trigger a callback using the `active` prop with the current version of  dmc (.14.4.0)

You can see a live minimal example here:  https://py.cafe/amward/dash-mantine-stepper-guide

If you meant a different use case, can you let us know?

AnnMarieW:
Looks like this is stale - closing for now.

PR #273, opened 2024-05-22
Fix datetime format that was messing with PM time values

RenaudLN:
Formatting with `hh` makes the PM times go back to AM in the datetimepicker.


snehilvj:
Thanks @RenaudLN  for this change.

PR #286, opened 2024-07-22
fix(import): add missing import to README example

PeriniM:
**Issue**: The following line throw a missing import error.

```python
``` 

**Solution**: Fixed missing `import dash` statement to README example.


AnnMarieW:
Nice catch - Thanks! :woman_dancing:

PR #301, opened 2024-09-01
fix for #291

### LINKED ISSUE #260: Multiselect doesn't update pillsList dynamically
### Describe the bug

Hi!

There is any option with "combobox" or "css" (or anything) to "refresh" the pillsList values on Multiselect?

image

The logic is:

From top to botton, if i change one value in a previus multiselect, this update the next multiselect.
The arrays/list coming to the next multiselect has the values updates, but the "input boxes (who has the x for close) doesn't disappear"

Exampled

```
First MultiSelect:
val1: A
val2: B
```

```
Second MultiSelect:
v1: A1
v2: A2
v3: B1
v4: B2
```

if I remove "A" from First MultiSelect this gonna remove all "A*" on the second Multiselect:

```
Second MultiSelect:
v3: B1
v4: B2

```
In my code, the list incoming to update the Second Multiselect is [B1, B2], so all my code works as i spect.

But, for some reason Mantine Multiselect does not update the boxes inside and keeps the previous view.

```
Second MultiSelect:
v1: A1
v2: A2
v3: B1
v4: B2
```
This "A1, A2" is just a view, if i close his boxes, just disappear and don't change anything
There is any css for "autorefresh" this pillsList?

### LINKED ISSUE #291: Broken dmc.MultiSelect in 0.14.4
Hello. Before we start, I would like to thank the developer for porting such a beautiful library to the Dash platform.

The problem is the "broken" dmc.MultiSelect component in version 0.14.4. 

The design code is:
``` python
dmc.MultiSelect(
w="47%",
searchable=True,
hidePickedOptions=True,
clearable=True,
label="Types to search",
id="n_search_in_types",
placeholder="Search all types",
disabled=True,
value=[])
```

Callback code
``` python
@callback(
    Output("n_search_in_types", "data"),
    Output("n_search_in_types", "disabled"),
    Output("n_search_in_category", "placeholder"),
    Output("n_search_in_types", "placeholder"),
    Input("n_search_in_category", "value"),
    State("n_search_in_types", "value"), # or Input, doesn't matter
)
def add_types_in_search(category_id, selected_types):
# some work code
return (
            types_select_data, # dict with values
            False,
            "Search by categories",
            (
                "Search by selected types"
                if len(selected_types) > 0
                else "Search by all types"
            ),
        )
```
The logic of the callback is to get the value of the `n_search_in_category` field from the user and suggest suitable types corresponding to the given category. 
In version 0.14.3 this worked as follows: get the value from `dmc.MultiSelect(id='n_search_in_category')` and add the required data to `dmc.MultiSelect(id='n_search_in_types')`.

In version 0.14.4 this stopped working, because now when `dmc.MultiSelect(id='n_search_in_types')` is updated (update data, not user values) - the selected parameters are reset. Moreover, even without updating this component and selecting some options in this component - the data is reset. Moreover, this does not happen in the component with categories.

### LINKED ISSUE #304: Cannot read properties of undefined (reading 'length') - dmc.MultiSelect data
I keep getting Cannot read properties of undefined (reading 'length') when running this callback, this didn’t happen to me on mantine v0.12 

It seems to happen when the data state is changed but not the value

```
import dash
from dash import dcc, html
from dash.dependencies import Input, Output

app = dash.Dash(__name__)

    children=[
        html.Div([
            dmc.MultiSelect(
                placeholder="Parameter",
                id="multi-select",
                data = ['Option 1', 'Option 2'],
                searchable=True,
                disabled = False,
                clearable=True,
            ),
            dcc.Input(id='input', value=''),
            html.Button('Update Options', id='update-button', n_clicks=0)
        ])
    ],
)

@app.callback(
    Output('multi-select', 'data'),
    Input('update-button', 'n_clicks'),
    Input('input', 'value')
)
def update_multiselect_data(n_clicks, new_option):
    if n_clicks > 0 and new_option:
        return ['Option 1', 'Option 2', 'Option 3', new_option]
    return ['Option 1', 'Option 2', 'Option 3']

if __name__ == '__main__':
    app.run_server(debug=True)
```

AnnMarieW:
fixes #291 
fixes #260 
fixes #304 

This PR ensures that the values in a dropdown are set correctly when the `data` prop is updated in a callback.

Checked that it worked with all valid formats for the `data` prop:

    1.  an array of strings 
    2.  an array of dicts with label and value properties.
    3.  an array of dict with group and items as keys where items are one of the previous two types.

### DIFF
--- a/pr-examples/pr-291.py
+++ b/pr-examples/pr-291.py
@@ -43,7 +43,7 @@ app.layout = dmc.MantineProvider(
                 label="Select Category",
                 id="n_search_in_category",
                 data=["A", "B", "C", "D"],
-                value=[],
+            #    value=[],   Checking for value is none closes PR 
             ),
             dmc.Text("STEP 2 Verify that all the dropdowns below update the values correctly regardless of the format of the `data`.", my=20),
             dmc.MultiSelect(
@@ -53,7 +53,7 @@ app.layout = dmc.MantineProvider(
                 clearable=True,
                 label="1. Data is array of strings",
                 id="n_search_in_types",
-                value=[],
+               # value=[],
                 mb=60,
             ),
             dmc.MultiSelect(
@@ -63,7 +63,7 @@ app.layout = dmc.MantineProvider(
                 clearable=True,
                 label="2. Data is dict with labels and values",
                 id="n_search_in_types_dict",
-                value=[],
+              #  value=[],
                 mb=60,
             ),
             dmc.MultiSelect(
@@ -73,7 +73,7 @@ app.layout = dmc.MantineProvider(
                 clearable=True,
                 label="3a. Data is dict with group and items keys, where items is array of strings",
                 id="n_search_in_types_groups",
-                value=[],
+             #   value=[],
                 mb=60,
             ),
             dmc.MultiSelect(
@@ -83,7 +83,7 @@ app.layout = dmc.MantineProvider(
                 clearable=True,
                 label="3b. Data is dict with group and items keys, where items is a dict with labels and values",
                 id="n_search_in_types_groups_3b",
-                value=[],
+              #  value=[],
                 mb=60,
             ),
         ], mb=200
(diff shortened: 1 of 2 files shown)


snehilvj:
Thanks a lot Ann!

PR #307, opened 2024-09-09
delete pr examples folders

AnnMarieW:
Seemed like a good idea, but had a change of heart :slightly_smiling_face:


AnnMarieW:
I'm going to merge these because it doesn't effect the code base.

PR #265, opened 2024-05-16
Solve #264 and #203

albavilanova:
Explained in #264

PR #314, opened 2024-09-14
Fix DateInput

### LINKED ISSUE #203: [BUG] The localization is not applied when rendering the component for the first time.
Hello

My code
```
import dash_bootstrap_components as dbc
from dash import Dash, html, dcc, Input, Output, State, no_update, clientside_callback, ClientsideFunction, callback_context

app = Dash(update_title=None)


app.layout = html.Div(
    children=[
        html.Div(
            className='containerDate',
            children=[
                dmc.DateRangePicker(
                    id="date-range-picker",
                    # label="Дата",
                    # description="You can also provide a description",
                    locale="ru",
                    persistence=True,
                    persistence_type='local'

                ),
                dmc.Space(h=10),
                dmc.Text(id="selected-date-date-range-picker"),
            ]
        )
    ]
)

if __name__ == '__main__':
    app.run(debug=True)
    
```

dash==2.12.1
Flask==2.2.5
dash-mantine-components==0.12.1

Problem with localization in case "persistence=True'". When opening the dmc.DateRangePicker the user sees locale="ru"(all values are in Russian).
But with the subsequent update, the language becomes eng

GIF
anim4

### LINKED ISSUE #249: DateInput does not use valueFormat setting
HI,
With the DateInput component, by setting valueFormat="DD/MM/YYYY" if I enter the date via keyboard it is not inserted according to the DD/MM/YYYY format. MM/DD/YYYY is entered. Everything is ok if I select the date with the combobox
for example if I write 09/26/1972 and press enter the field will write 02/01/2001. If I write 02/03/2020, in the field it says 03/02/2020
I wrote the same for DatePicker (https://github.com/snehilvj/dash-mantine-components/issues/131); it was a bug in Mantine v5 at the time, now I think Mantine v7 has the same bug, avoidable by using the dateParser prop to replace the default date parser.
is there a solution for dash mantine?

### LINKED ISSUE #264: Free input on DateInput component does not consider locale
Hi,

I've noticed that whenever I want to set the date of the DateInput component manually, the input is only accepted if it is passed in English, ignoring the locale previously set. Turns out that the date is considered invalid because there is no extension of dayjs to custom formats.

I will soon send a pull request to solve this issue, as well as #203. For that one, what's happening is that the initial state of the date is set in the DateInput (using setState) before the locale is set in DateProvider (using useEffect). Removing useEffect and defining the component like this solve both issues:

```
import {
    DatesProviderProps,
    DatesProvider as MantineDatesProvider,
} from "@mantine/dates";
import dayjs from "dayjs";
import React, { useEffect } from "react";
import customParseFormat from 'dayjs/plugin/customParseFormat';

interface Props extends DatesProviderProps {
    /** Unique ID to identify this component in Dash callbacks. */
    id?: string;
    /** Update props to trigger callbacks. */
    setProps: (props: Record<string, any>) => void;
}

/** DatesProvider */
const DatesProvider = (props: Props) => {
    const { settings, children, setProps, ...others } = props;
    const { locale } = settings;
    
    const localeObject = window[`dayjs_locale_${locale}`];
    if (localeObject) {
        dayjs.locale(locale, localeObject);
        dayjs.extend(customParseFormat);
    }

    return (
        <MantineDatesProvider settings={settings} {...others}>
            {children}
        </MantineDatesProvider>
    );
};

DatesProvider.defaultProps = {};

export default DatesProvider;
```

Feel free to update my pull request, I am not familiar with the codebase and this might have an effect on something else, you will know better.

Best, 

Alba

### LINKED ISSUE #308: DateInput Date Time mode does not register the time
The component returns only the date. Can replicate issue on website/using website example

```
from datetime import datetime

from dash import Output, callback, Input, html

html.Div(
    [
        dmc.DateInput(
            id="datetime-time",
            valueFormat="DD/MM/YYYY HH:mm:ss",
            label="Date and Time input",
            value=datetime.now(),
            w=250,
        ),
        dmc.Space(h=10),
        dmc.Text(id="selected-datetime"),
    ]
)


@callback(Output("selected-datetime", "children"), Input("datetime-time", "value"))
def update_output(d):
    prefix = "You entered: "
    if d:
        return prefix + d
    else:
        return ""

```

AnnMarieW:
fixes #308
fixes #249
fixes #203
fixes #264

This PR is based on PR #265 by  @[albavilanova](https://github.com/snehilvj/dash-mantine-components/issues?q=is%3Apr+is%3Aopen+author%3Aalbavilanova)

@[albavilanova](https://github.com/snehilvj/dash-mantine-components/issues?q=is%3Apr+is%3Aopen+author%3Aalbavilanova)  Thanks so much for your PR.  This new one is based on your work and includes a few sample apps for testing.  I was reluctant to update your PR because I understand you are using your own version of dmc in production.   If you would like to continue to work on this, let me know and I'll close this PR in favor of yours.   Otherwise we can carry on with this one and we'll give you credit in the Changelog when it's merged.

This PR also enables the date parsing  to work with the `DateInput` component even when the `DateProvider` is not being used. 

@snehil - was there any reason not to merge the original PR 265?  

Note - The three sample apps for testing are in the `usage-265` folder and can be deleted after merging.


alexcjohnson:
Two usage questions:
1. Does it work to have two different locales on different components?
2. Does it work to change the locale of an component via callback?

AnnMarieW:
The issue was that the locale wasn't working when the component first rendered.  I'll try your solution and see if it works - I'll also verify with the two usage items too.    Thanks for the suggestions :slightly_smiling_face:

AnnMarieW:
@alexcjohnson This is now ready - and verified that additional use-cases work :tada:

albavilanova:
Thanks @AnnMarieW for working on this, I just used my own version because of this issue, but I will be back to using this one once it is updated in PyPI.

alexcjohnson (approved):
Fantastic! 💃

PR #316, opened 2024-09-20
Added `autoplay` option to Carousel component

mmarfat:
### Description
This pull request introduces a new feature to the `Carousel` component. It allows users to enable and/or customize the behavior of the autoplay feature by passing a set of options that control the autoplay functionality.

### Key Changes:

- **Added `AutoplayOptions` Interface**:
  - Created a new file `autoplayOptions.ts` to define the `AutoplayOptions` interface. This interface includes various options to customize autoplay behavior:
    - `delay`: Number of milliseconds between transitions (default: 2000).
    - `jump`: If true, transitions will be instant when advancing (default: `false`).
    - `playOnInit`: If false, autoplay will not start automatically (default: `true`).
    - `stopOnInteraction`: If false, autoplay will not stop after drag interactions (default: `true`).
    - `stopOnMouseEnter`: If true, autoplay will stop when the mouse enters the carousel (default: `false`).
    - `stopOnFocusIn`: If true, autoplay will stop when a focusable element inside the carousel receives focus (default: `true`).
    - `stopOnLastSnap`: If true, autoplay will stop when reaching the last slide (default: `false`).
    - `rootNode`: A custom node that responds to user interactions (default: `null`).

- **Updated `Carousel` Component**:
  - Modified the `Carousel` component to accept the `autoplay` prop, which can now be either a boolean or an object containing the `AutoplayOptions`.
  - Integrated the `AutoplayOptions` into the autoplay logic of the `Carousel`.
  
### Example usage:

An example has been provided in the [docs](https://github.com/snehilvj/dmc-docs/pull/69).


AnnMarieW:
Hi @mmarfat 

Thanks for this PR and for adding the docs too! :rocket:   Autoplay will be a nice addition to the `dmc.Carousel`

I tried an example with the following and it stopped on mouse enter, but didn't resume after.  Does it work for you?
`autoplay={"delay": 2000, "stopOnMouseEnter": True}`

mmarfat:
Hi @AnnMarieW 

Yes, from the [Embla Carousel docs](https://www.embla-carousel.com/plugins/autoplay/#stoponmouseenter) it seems that this is the default behavior. For the autoplay to resume after entering and then leaving the carousel the flag `stopOnInteraction` has to be set to `false` (by default it's set to `true`).

> ### stopOnMouseEnter 
> Type: `boolean`
> Default: `false`
>
> When enabled, autoplay will stop when a mouse pointer enters the Embla Carousel container. If [stopOnInteraction](https://www.embla-carousel.com/plugins/autoplay/#stoponinteraction) is also false, autoplay will resume when the mouse leaves the carousel container.

AnnMarieW:
Oh, yes - I see. This works now:
```
autoplay={"delay": 2000, "stopOnMouseEnter": True, "stopOnInteraction":False}
```

I  removed the following and the autoplay paused and resumed as expected.  Any reason to include?

```
            onMouseEnter={
                autoplay && autoplayOptions.stopOnMouseEnter
                    ? autoplayRef.current.stop
                    : undefined
            }
            onMouseLeave={
                autoplay && autoplayOptions.stopOnMouseEnter
                    ? autoplayRef.current.reset
                    : undefined
            }

```


Since you are not changing the defaults from the upstream Embla defaults, it may not be necessary to set them here.  Also, it may not be necessary to define the types in the `autoplyOptions.tx` file.  That would be consistent with how it works in other components like when props are passed to `recharts`.   But best to get feedback from @snehilvj  on this one.

mmarfat:
Yes, I agree. Most of it was defined just to be as explicit as possible but I can see how it's a bit redundant. Feel free to modify and/or remove any code you deem unnecessary or not in style of the other components.

AnnMarieW:
Thanks for the update @mmarfat .  Just need to wait for @snehilvj 's review.  He's been working on some changes with the props, so he can chime in on how he would like this set up.

Snehil, Just to save you some time for the review, here is a little sample app to run.  In this version if you hover over the slide, it stops, then resumes on exit.  If you change the `autoplay` to True, it runs with the Embla defaults,

```python 
from dash import Dash, _dash_renderer, html

component = dmc.Carousel(
    [
        dmc.CarouselSlide(dmc.Center("Slide-1", bg="blue", c="white", p=60)),
        dmc.CarouselSlide(dmc.Center("Slide-2", bg="blue", c="white", p=60)),
        dmc.CarouselSlide(dmc.Center("Slide-3", bg="blue", c="white", p=60)),
    ],
    id="carousel-simple",
    loop=True,
    autoplay={"delay": 2000, "stopOnMouseEnter": True, "stopOnInteraction":False},
    #autoplay=True,
)


    html.Div([component])

)

if __name__ == "__main__":
    app.run(debug=True)


```

AnnMarieW:
Hi @mmarfat 

Just getting back to this - thanks for your patience.

I think the way the props are passed to the Embla component is fine.  

One issue I found is that the carousel cannot be started or changed in a callback.  Here is a sample app

```python
from dash import Dash, _dash_renderer, html, Input, Output

component = dmc.Carousel(
    [
        dmc.CarouselSlide(dmc.Center("Slide-1", bg="blue", c="white", p=60)),
        dmc.CarouselSlide(dmc.Center("Slide-2", bg="blue", c="white", p=60)),
        dmc.CarouselSlide(dmc.Center("Slide-3", bg="blue", c="white", p=60)),
    ],
    id="carousel",
    loop=True,
    #autoplay={"delay": 2000, "stopOnMouseEnter": True, "stopOnInteraction":False},
    #autoplay=True,
)


    html.Div([
        dmc.Button("update autoplay", id="btn", n_clicks=0, m=10),
        component
    ])
)

@app.callback(
    Output("carousel", "autoplay"),
    Input("btn", "n_clicks")
)
def update(n):
    print(n)
    if n % 2 == 0:
        return True
    return {"delay": 2000, "stopOnMouseEnter": True, "stopOnInteraction":False}

if __name__ == "__main__":
    app.run(debug=True)

```

mmarfat:
Hi @AnnMarieW 

No problem, better to be patient and make sure it works properly 😎.

You're right that it didn't work which surprised me since I tested it in the original commit and it worked fine. The reason it stopped working is because I used `useRef()` instead of just directly handling the autoplay prop (JS/TS aren't my main prog. languages so I just followed the example from the [Mantine](https://mantine.dev/x/carousel/#embla-plugins) website).

I've tested it now with your sample app (thanks for that) and it seems to be working fine. Let me know if you run into more issues.

AnnMarieW:
That seems to work.  Last comments before final review:

- don't need the `{ delay: 4000 }` since that is the default.
- update with the current upstream master branch
- needs an entry into CHANGELOG.md

At the top of the file, you can add something like

```
# 0.14.6 UNRELEASED

### Added

- added the `autoplay` prop to the Carousel #316 by @mmarfat 

```

AnnMarieW:
Note - @alexcjohnson  - the upstream docs for this is here:  https://mantine.dev/x/carousel/#embla-plugins

mmarfat:
Okay, everything should be there now. Let me know if anything else is needed.

package.json
```diff
@@ -55,7 +55,8 @@
         "embla-carousel-react": "^7.1.0",
         "is-absolute-url": "^4.0.1",
         "jsonpath": "^1.1.1",
-        "recharts": "^2.12.0"
+        "recharts": "^2.12.0",
+	"embla-carousel-autoplay": "^7.1.0"
```
alexcjohnson, line 59: nit: let's insert this alphabetically. But also, shouldn't there be a corresponding `package-lock.json` update?

src/ts/components/extensions/carousel/Carousel.tsx
```diff
@@ -1,66 +1,76 @@
...
+}
+
+/** Carousel */
+const Carousel = (props: Props) => {
+  const { children, setProps, autoplay, ...others } = props;
+
+  const autoplayPlugin =
+    autoplay === true
+      ? Autoplay()
+      : typeof autoplay === "object"
```
alexcjohnson, line 67:
```suggestion
      : autoplay && typeof autoplay === "object"
```
just in case we get `autoplay=null` somehow (particularly because Dash isn't yet good at ts typechecking).

alexcjohnson:
Looks great, just had two small comments. Thanks for adding this @mmarfat!

mmarfat:
Thanks for the review @alexcjohnson. I've added the requested changes so hopefully everything is ready now.

alexcjohnson (approved):
💃 Thanks @mmarfat, Looks good now!

AnnMarieW:
@mmarfat - Thank you for your first PR to dmc 🎉 We really appreciate your contribution 

Thanks @alexcjohnson for reviewing :pray:

PR #317, opened 2024-09-22
updating dependencies and readme

AnnMarieW:
This PR updates the dependencies to lock the exact version of the upstream Mantine version and the  optional CSS files. 

Previously the Mantine dependencies were locked to the major version and they could get out of sync with the CSS files which specified an  exact version.

Updated the README.md with a few more tips for contributing.


AnnMarieW:
Note - the sample app is a test to make sure the Notifications work with the updated dependencies.  

Previously, with a new build, the Notifications did not display because the stylesheet and the Mantine versions did not match.

alexcjohnson (approved):
💃 Yep, makes sense!

PR #325, opened 2024-10-01
fix loading state props

AnnMarieW:
Currently the `loading_state` is included in the dom for every dmc component. For example

image

This PR makes it so the `loading_state` is used as intended and is consistent with all other component libraries such as html, dcc and dbc by setting:

```
           data-dash-is-loading={
                (loading_state && loading_state.is_loading) || undefined
            }

```

As a sample I just updated the `dbc.Alert`.  If you run the basic sample app you will see the `loading_state="[object Object]"` is no longer included.

```python

    dmc.Alert(
       "Hi from Dash Mantine Components. You can create some great looking dashboards using me!",
       title="Welcome!",
       color="violet",
    )
)

if __name__ == "__main__":
    app.run(debug=True)

```

Hey @alexcjohnson - can you give this a quick look and a "pre-approval" before I go ahead make this same change to like 100 components?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 
 - Added `autoplay` prop to the `Carousel` component #316 by @mmarfat 
 
+### Fixed
+
+- Excluded the `loading_state` prop from being passed to the DOM.   Added `data-dash-is-loading` attribute to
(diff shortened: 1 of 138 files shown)


alexcjohnson:
Yep, makes sense to me, good catch!

AnnMarieW:
ahhh, so much better now :slightly_smiling_face: 

image

PR #347, opened 2024-10-11
minify metadata.py to reduce package size

### LINKED ISSUE #271: Reducing package size by trimming metadata.json
When deploying python tools to cloud servers, package size is relevant to consider. When inspecting the package contents for dash mantine components (Linux version of the package, as the cloud runs on Linux), I see that the package is about 9MB. The major contributor to that is `metadata.json`. When opening I see that the file is constructed in nice format for humans, with a lot of indents. 

Is it possible to trim that, doing json.dumps with no indent specified? A small test on my system allowed me to reduce the size of the file by 2MB, which is about 20-25% size reduction.

AnnMarieW:
closes #271 

This PR reduces the size of the metadata.json from 5.5MB to 3.0MB :tada: 

- adds a new script `minify_metadata.py` to reduce the file size
- updates the `dist` script to run  `minify_metadata.py` before creating the source distribution 
- updates the `release.md` file so I remember how do to this when making a release :stuck_out_tongue:

### DIFF
--- a/.github/releasing.md
+++ b/.github/releasing.md
@@ -18,9 +18,9 @@ On the Master branch:
 $ rm -rf dist
 ```
 
-2. Build source distribution
+2. Build source distribution.  Note this will also minify the metadata.json file which reduces the package size
 ```
- $ python3 -m build --sdist --wheel
+npm run dist
 
 ```
 3. Test your tarball by copying it into a new environment and installing it locally, for example:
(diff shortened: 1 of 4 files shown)


alexcjohnson (approved):
💃 But then if https://github.com/plotly/dash/pull/3034 is merged and released we can remove this step here

BSd3v:
@alexcjohnson, this would introduce a dependency on the dash version for building?

alexcjohnson:
Yes, but since that only applies to the one dev making the release it shouldn’t be a problem

PR #352, opened 2024-10-12
Fix: enable stepping when clicking on steps buttons

RenaudLN:
dmc-stepper-step-click

### DIFF
--- a/tests/test_stepper.py
+++ b/tests/test_stepper.py
@@ -1,0 +1,41 @@
+from dash import Dash, html, Output, Input, _dash_renderer
+import dash_mantine_components as dmc
+
+_dash_renderer._set_react_version("18.2.0")
+
+
+def test_001_st_stepper(dash_duo):
+    app = Dash(__name__)
+
+    app.layout = dmc.MantineProvider(
+        html.Div([
+            dmc.Stepper(
+                id="stepper",
+                active=0,
+                children=[
+                    dmc.StepperStep(children="Step 1", label="Step 1"),
+                    dmc.StepperStep(children="Step 2", label="Step 2"),
+                    dmc.StepperStep(children="Step 3", label="Step 3"),
+                ],
+            ),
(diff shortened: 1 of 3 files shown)


AnnMarieW:
@RenaudLN  - This looks fantastic!  Thanks so much :trophy: 

@alexcjohnson any comments before merging?

tests/test_stepper.py
```diff
@@ -0,0 +1,66 @@
...
+        Output("stepper", "active", allow_duplicate=True),
+        Input("next-button", "n_clicks"),
+        State("stepper", "active"),
+        prevent_initial_call=True,
+    )
+
+    dash_duo.start_server(app)
+
+    # Wait for the app to load
+    dash_duo.wait_for_element("#stepper")
```
alexcjohnson, line 54: Tiny point: it's a bit better (both in terms of ensuring the app is fully loaded and for understanding what the test is doing) to wait for the initial condition of the output you care about, here the `div.mantine-Stepper-content`, which I guess is `"Step 1"` since you're starting at `active=0`.

If that's the case, then the first button click below is a noop - which may actually be a worthwhile thing to test, I just didn't realize that's what it was doing until I figured out the initial condition.

And in the interest of completeness, it's good to ensure no JS errors were thrown by the test, by calling:
[`assert dash_duo.get_logs() == []`](https://github.com/plotly/dash/blob/a354a735332af0f14be941baf4c1a8ea133ec133/tests/integration/callbacks/test_basic_callback.py#L66)
at the very end of the test. (That form is better than the `assert not dash_duo.get_logs()` that we have in some places; they test the same thing but `== []` gives more useful messages when it fails)

alexcjohnson (approved):
💃 Looks great, thanks @RenaudLN, for tracking this down and for including a test! Just had one non-blocking comment there.

AnnMarieW:
Thanks again @RenaudLN  :woman_dancing:

PR #366, opened 2024-10-19
Reduce Highlight file size

AnnMarieW:
This PR reduces the file size of `Highlight.py` from  442.9kB to 10.8kB

Previously, the file was over 7000 lines because it defined every possible CSS variable that could be used within the `HighlightStyles` prop.

### DIFF
--- a/src/ts/components/core/Highlight.tsx
+++ b/src/ts/components/core/Highlight.tsx
@@ -8,7 +8,7 @@ interface Props extends DashBaseProps, TextProps {
     highlight: string | string[];
     /** Key of `theme.colors` or any valid CSS color, passed to `Mark` component `color` prop, `yellow` by default */
     color?: MantineColor | string;
-    /** Styles applied to `mark` elements */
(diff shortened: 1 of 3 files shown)


src/ts/components/core/Highlight.tsx
```diff
@@ -9,7 +9,7 @@ interface Props extends DashBaseProps, TextProps {
     /** Key of `theme.colors` or any valid CSS color, passed to `Mark` component `color` prop, `yellow` by default */
     color?: MantineColor | string;
     /** Styles applied to `mark` elements */
-    highlightStyles?: React.CSSProperties;
```
alexcjohnson, line 12: Oh wow, I would not have guessed that `React.CSSProperties` blows up into a whole CSS reference manual! (cc @t4rk1n I wonder if we could simplify this particular case at the codegen stage?)

This change looks great, I'd just suggest adding a note that these are CSS but property names are camelCased - that's pretty universal across Dash but still trips people up.

alexcjohnson (approved):
💃

PR #368, opened 2024-10-20
Additional chart events + new charts

michaelw88:
Hi, 

this is my first PR on github, so please bare with me if I did something wrong. 

Changes:
- Added Composite and Bubble chart support.
- Added hoverData prop to all chart types (basically, includes same information as normal click event but triggers on mouse over instead of click)
- Added clickedSeriesName and hoverSeriesName: When hovering or clicking a chart Series, this includes the series name. Implemented for: BarChart, AreaChart, LineChart, CompositeChart
- Added highlighting of hovered series: Works like hovering of a chart legend entry, i.e. all not-hovered series gets reduced in Opacity. Can be enabled/disabled by setting highlightHover in chart props. Works for BarChart, AreaChart, LineChart, DonutChart, PieChart,CompositeChart

Added a sample app "tests/chart_test.py" to test all changes done.


AnnMarieW:
Hi @michaelw88 

Awesome PR for your first one!  :rocket: 

To start, could you update with the upstream master and see if you can resolve the conflicts?

michaelw88:
Hi @AnnMarieW ,

thanks - I resolved them.

AnnMarieW:
Thanks for the update!
After the merge, did you try rebuilding and running the test app?  
It looks like there are some duplicate and missing props.

AnnMarieW:
@michaelw88 


This looks really good!  I appreciate the addition of the  BubbleChart and the CompositeChart too 🎉 


I’m still doing the review, so I’ll likely have more comments later. Just wanted to share a few thoughts so far:

- It would be better to default the `highlightHover` to False otherwise it might be considered a  a breaking change.

- In the `PieChart`, it looks like it’s no longer possible to pass additional props using  `pieProps`. For example the animation example in the docs on the PieCharts page doesn’t work with the build from this branch.  (Other chart types annimation example works ok)

-  It’s more efficient in Dash to minimize the number of calls to `setProps`.  For some charts,  `setProps` is called twice for each click, once for `clickData` and again for `clickSeriesName`.  Is there a way to combine the click handlers?

michaelw88:
@AnnMarieW 

Thanks.

- Re combination of click handlers. Possible, the click event on the chart is the parent so always triggered after a click on a series. With that, I am using `const seriesName = useRef(null)` to store the `clickSeriesName` temporarily in the click handler on the series and use it to update the props in the main click event.  Same for `hoverSeriesName`
- I also realized that when a series is clicked and then a click outside of the series is done, the prop `clickSeriesName` is not updated. This might not be the ideal behavior, so I changed this to update the `clickSeriesName` back to null. Same for `hoverSeriesName`
- Pie and Donut: The color logic used by Mantine is unfortunately different than for other chart types (i.e., color is defined on `data` rather than `series`). Trying to still highlight hover event seems to have killed all chart animations, hence I took the highlight capability out for now, as I have no idea if highlighting for Pie and Donut charts is even possible without changing something in  Mantine.
- `highlightHover `defaults to `False `now.

AnnMarieW:
@michaelw88  
Nice solution for combining the click and hover handlers :star_struck: 

Good pick removing the highlight hover for Pie and Donut for now.  We could revisit in a future PR.

@alexcjohnson do you have any comments before we start on tests?  Do you think there will be any performance issues with including the hover data feature?

alexcjohnson:
@michaelw88 very impressive for a first PR! One small procedural matter that will help reviewing: in whatever editor you're using, there should be a setting to trim trailing whitespace on save. If you turn that on, all the changed lines will represent real edits you've made.

Hover data should not be a problem for performance. Various other components (including `dcc.Graph`) do similarly high-rate updates, and these only cause problems if you hook them up to slow server-side callbacks (and even then it's generally only a subpar experience, not really a problem 😅).

AnnMarieW:
Here is the sample app hosted on PyCafe - based on the unreleased dmc version in the branch:
[PyCafe Sample App](https://py.cafe/snippet/dash/v1#c=H4sIAD8NGGcAA-1bbW_bOBL-K4IOh6aA41hy4iQGBGya3m53L70rrr0XIA4MWqZjNbLkk-QkbpH_fs-QlETJMi3nUmw-rNM0Jjl8ODMczpBD6bvtx1NuD-1ZEi-sKUvnVrBYxklm-SwMJ8y_61hT3-9Yv0bLVdax_r7KxN95tgg71seLL5cfOtZ7dOtYY-o9Tng05QlPRpFATFg0xR-FSaUgykbRKFI1os-CRVkQ8bEfozLiUZZaLLWmC58IR1EVuDtOeYYS87PxPU_SII4ORrZz1nW7vZH9ljpMWcYsz7oeRRY-30f2Io6y-cgeWiP7NxatWLIe2R0UPi9Yki3nGDOlVsft9aj6ii2zeCmqzmXNFzYJeSZqQPPUaUL-mU-S7dAKSIPOR9Owj7dhf2SJj68NwOhSx-215_limQRhM8MKRQPe5PdsO7_NakCHOrdKAA2VFNMM-9sq4o24pyd13MEmLCR6GkU3ZCGT1QT1MKzSUL6Deh6vEiJ90zsb9npvCCCA1T2K7lS6Z-EKHKB00rMkj5V-Tm9Xv8GgsZ-7q99p83jHz-RzsKMfpqCx3y69YDab-rk79bKl3069DGQ_bVYv5zANzCr8R_ddWXMgDWruDXrKtoKpB_PK1kuBNLJ9IpPmVYw2siXoyM4tkmzG0-xHVcO13XLv2hl0LNc9uVG1IZvw0IO5spCnR1Is1eTHYZygKQwWvDsoqgnyr3ztkRoEB2-ol1ABlhSKgjVR_ibKQhVvLFi28H0TllQUoIpK-pYiM7CZyzv3-lhKVeaKRanrRFPGQ5DNr_gtPLb3JVnxDqoUWHA7D_GbfYjhu1WjbMKgnxIsXmIwSC8QIFgG337hZ8G94JaIc6ZIAnCRZghQfFrwkfIk4KmnHL_yHxFbKGkrToNkFnMg2-6DGI5CTIRQBsILxr3cIHBKzWzgF-6nhj3BDBmQRbMJt3BiNdyMs9CAK5o1XDJKYSM_5bFdWYUM6gftLQPfaLYRbtXGABF4kkWHaBb9Dlk-fRIhGvth4N9BAoyPf1M-s1ZLQPCD6O1QLSCerZIoD9pSB8VX-miRQHzRtCWbazFBbTYOKDTAhN9u0GvRIqftgxaub5NWiyM5rQtauFedFosw_zqLE8mmFZBQ1X1HdaugBXctHhchVAY9ODgxh3JLNIoYdkD6Qr_Iy3utdEL5_5b6H0tu15Kjv_4quedfpNMKsdWlZZS7MiyMK1Sh5ZFmvPSgvyRsGmA37P3MwpRrDf-5eAzSjdr3cbZZqXthtfppfN1yaPD9LYdQXjZI_GFPe7jw1GdZxpN8B0t_XP3IU3CxKcVJMSOyvWTylyReLS0MV2mXnn6oe2YlHruVe8UT4v7dx1_lxlHXQZWQvGtJ6BoIdUTnbDvhcQXxxECoI_YNPLpn-tAGRNfVCQ3CyKNJzqNBmH4F8dwgjM5j3zHwiN1oOfSxYWgd0T01DK3z6Bp4dM91YQaGoY91xL5haJ2wbyB0-_rQJqlP9aENPB7rwvQNiK6O6Jj0KI4xOaJpKeiErmGuj_W57psMV9ejY9BjX9ejcbnq6jFZ-LGuHtMq7FcM17SuKzya7FHXo-NoOzW1syozDs2-M-HTna7T3dt1VuQ0zG7VDNratNGJVFaywVQrK9loqs_z2qYl39prV6Q22EvVpk0WWPGIL2GBlehn9IgVozAovOJEjOuu4htMrrMS_QxGUfXarX2DycdW3LtpKVTCpCmeVqbQZGaVKTTpsRJ4TT62shRMy7UyhS_iESuRxbRmahF60yOqtJraa-pnhs9aVZFYK48A7Y4QCreWWtN3tvWcmEqJKRGKnJiSIod5pOPRlcq8XQhK2bDWG0SXfDtd_lBq4hYXDDTmgcitpaYMhRYCLiRHRWpSdr7u3dS2_NBAcBvTpn_raeFdM5RTh1rzMIwfjFCXzVBuHSo_iGwFet8M1K8D3SZsrQMJE5rG0SrTDeh9UbHXqVPglNjLgLfPGQrrKqb2GhnnjqX-3ci0FP0AUufzkyruxSUwZJbsxzD5_PRdoT4tgScx9CyeoHp2Hi-YWZH1ZwunUc_qqTpt9ZSyiYsf8hnyrgaikoSNpHU1yJ_n66GYIIMWQPOqdUBEo0hcmaZBVhitsNrLSu2B7v-LVHw7Uy7wn5vrqY2tMfBimZ5ShsrFxbNyOyVWNcH1rIROCVbNs4osDn7zeOOjNcP1t5gs8f-YupI1FfZUVpNRKVE10xLtZQao-MhbsCwTt-T_FHZqCasgFmED2xP39SQ43dDInvWNAg8bGFS-ZjeL7Tjc5pTy5ehhXSwmZIi4DqqzXoafdrwL__BSnDc7klZ858GoHdfSYhvYLjK-hs75NWdDd-1a1QBQ7OYaEPQdpAFCczYNIFVXl8OkvIG0uByp-FIVJvKZ-yUJploVfeqLR6NEUrZGnH_SJYs8x62vloKlIJzi2RVP-tcNmo1V1n7MnUMSzhf-SBFNGNt7GexelAknDwW7uahHnFjEa7kOJGuV4FNax9PLsryP3uZ0T_069aax9vr0Jib1s4jwfxMx87Vpb4PB16dDMcGvWYcbDO6lw1pFlUTFOUEpv-IPWy7h3OlpwwNwxZOIheM0W-PRmjnHXkzwK8vdi6srcU5Al27I1mBYHeg-ymcNcSK7D_A8odwbq8EKwbU4IHTAJmm7SEGUV0GaHxfrn4Ze9d74xdTTbT6pUxyysTnI2GQs95AbetuiwO3YeCSoDi02kwK5NYjaLtRwVG2TVTbjFCeVOpS2F9gictshRKqhDq92pwq6LRRdldeR5K5rTyCkFeo4Ys-5fXb3Flul6OrDFNu0HUM1D6Mty_yzzQMRJ59YxMODyvEmPwth-Cbj3hMrN9wGY94XqTDd1hZtxKva71bL3hO1NNtGY94TrTDdJnveEys33waT3hNJt9CWlluLHbUR8yctcYNXuZ-jT4zgFWXiVAYKhDM8jsPoCawq2aapbsQnnRHBAdW3S1K1CqfiWXuRs8rDVD1vtTOdU0Jsbi1kJrxMZtW5OaBszlsg6yeqNEtkfSH-_gI37GBfXlS1g28Wkhp3ibenZObd5YvKt7nT3pCyRvIjZP3Rs6ifX5vlazeLuZUi-zAeUzZvPJa5hzHeQgmi8bhIPdDOMVlFB1M-Wd2KJ9QwqN2BF_nvKkj4gl5TwXsz9G4KvXHS_BrLT3hRJlumw6Oj5brrsxk_up0fQc5ghke9jtKIQzP3X4-o-6Hqflh2P3J75z130D8_HwiShhEOe13nuDs4XK77hxGqkGRadx_mIRgl1SoGUYI4_wr4gz2ciScEbT4Nsr9ElMS0hxk9fmcv1xmyreixXMdT7I0P73tdd9B10FnuoO3hd_KQ9MaNPXShiTjO_hED8ns-VoJSp5hue3hdtMBx4rUdND4E02xuD_HgbMdeBNG_ZRF341T6wOnZcFkMpvRWUhDyd0BFrvgSCWbMEE-2jECkhxNJC5IlI1zbfrp56mxysYXFsh_N_nKNjG1zs24E3ewxqw7TQg27RVei7BI7l7hZzrlCdU5P2oyJ8IpqFu4aNKejUennqSPUj_VwfQNrif07KhIKFAT6P1lzWOHQmgxY3z077c1m7sw_Pz8-d04Hp77T82cnbo_NTvD6ArYfd3gzwHO6eKML9xsUkWeB57k99xiveFEVHAWkg7knCxYG34i63-2LJnIEnnfWdbqnclGiYxfvhjmyhLWV6AuMWnt4bUw20jttTY2_0xJWXGEuQ-55J5KXWchSSNhHqQ8vNo0YFSCsfJ8uDCaHC54xcf8CPaAXmrJ0Sm-n4BElIRS9KBd9DaKvzKXODiHhsuVutUwhntAYnnSKcPGCnHW6jvwgpvkYULdlGGfhmtjBdECrua8gHXkeHAZV0upAZzFY36FucMXJOohuCaffPcYzBfDNuI2JQxANzsEnuqXBI7U7YiBYDvODDEOdS8lhegA4pIM_KOMIHcGBQF8lIQTvS9kA9MCTu28cXlvoCaN9C5bgDiYCBjFOCt3aT_8DExe-5vs4AAA)

AnnMarieW:
Hi @michaelw88 

The tests will pass if you rename the sample app `chart_test.py` to something that doesn't include the word "test".

Next step is to write tests.  If you are up for that, you can find some  info to help get started below. If not, just let me know if it's OK to push changes to your PR and I'll see what I can do.  Just need to test the new props that were added - quite similar to your sample app, but one file for each chart type.

- Info on dash testing:  https://dash.plotly.com/testing  

-  Example of testing  clickData with the LineChart.  This could be modified to include `clickSeriesName` as well.
https://github.com/snehilvj/dash-mantine-components/blob/master/tests/charts/test_line_chart.py

- The on hover might be a little tricker.  I haven't don that before, but I found an example in the dash repo here: 
https://github.com/plotly/dash/blob/0d9cd2c2a611e1b8cce21d1d46b69234c20bdb11/components/dash-core-components/tests/integration/tooltip/test_tooltip.py#L71

michaelw88:
Hi @AnnMarieW ,

thanks, I will look into it and try to push changes throughout the day. 
@alexcjohnson Thanks, re the info on trailing whitespaces. Did not find it by default in VS Code, but found an extension to do that. I will use it going foward.

AnnMarieW:
@michaelw88 
I'm working on the docs for the BubbleChart and it looks like a lot more props are included than are actually available for the BubbleChart.

https://mantine.dev/charts/bubble-chart/?t=props

What shows up in the docs are all the props that are available for charts like the LineChart.  You can see them in the docstrings of the .py files after a build,
`dash_mantine_components/BubbleChart.py`

Here's just a few that shouldn't be included?:
```

- withLegend (boolean; optional):
    Determines whether chart legend should be displayed, `False` by
    default.

- withRightYAxis (boolean; optional):
    Determines whether additional y-axis should be displayed on the
    right side of the chart, False by default.

- withTooltip (boolean; optional):
    Determines whether the tooltip should be displayed, `True` by
    default.

- withXAxis (boolean; optional):
    Determines whether x-axis should be hidden, `True` by default.

- withYAxis (boolean; optional):
    Determines whether y-axis should be hidden, `True` by default.

```

I haven't had time to dig in yet, but thought I'd give you the heads-up in case the answer is obvious to you :slightly_smiling_face:

michaelw88:
Hi @AnnMarieW ,

I incorrectly extended Props with GridChartBaseProps. Should be correct now for the bubble chart.

In terms of consistency, I also added click/hoverSeriesName to Donut, Pie and Scatter chart. Even though this information is in the click/hoverData, it might make it more user friendly to work with the same props.

I also added event handling of hover/click events for the dots of Area, Line and Composite charts. Without this, clicking or hovering over a dot would set clickSeriesName / hoverSeriesName to null. 

Regarding tests, I somehow didn´t make it work consistently (e.g., struggle to find the right element for Linecharts for Series events). If it´s straight forward for you, please feel free to push changes to the PR. Also, I renamed the sample app.

AnnMarieW:
@michaelw88 
Thanks for the updates - nice catch on the dot props, and good idea to add clickSeriesName to the other charts.

Before I attempt some tests, could you please update with the upstream Master?  I tried to do that locally and it had merge conflicts -- which was odd.  I don't want to mess up your PR :see_no_evil:

tests/charts_app.py
```diff
@@ -0,0 +1,365 @@
...
+    id={"type": "chart", "index": "line"},
+    h=300,
+    dataKey="month",
+    data=data,
+    withLegend=True,
+    series=[
+        {"name": "Smartphones", "color": "violet.6", "inactiveColor": "violet.1"},
+        {"name": "Laptops", "color": "blue.6", "inactiveColor": "blue.1"},
+        {"name": "Tablets", "color": "teal.6", "inactiveColor": "teal.1"},
+    ],
```
AnnMarieW, line 93: Found something weird... if you set `withTooltip=False` on the lineChart, then the click/hoverSeriesName is either null or None

AnnMarieW: Actually, when you specify `withTooltip=False`, the docs say that "It also removes the cursor line and disables interactions with the chart."   So that might be the issue?  If so, I could just note that in the docs.

https://mantine.dev/charts/line-chart/#remove-tooltip

michaelw88:
I updated with upstream master and pushed changes again. Does not seem to actually change anything though...

Re Tooltip: If set to false, Mantine is not rendering the Tooltip at all. As rechart requires the tooltip to calculate the ActivePoint of a chart, I suspect this to kill the event handling on dots: [Code](https://github.com/recharts/recharts/blob/11174ac611ae5b7ceb67f590af81dd3338ae828c/src/component/ActivePoints.tsx#L73). However, clicking on the line (instead of dots) still work for me.

We could do a workaround: When users define `withTooltip=False`, we could set it to `True` and instead adjust tooltip styling to be invisible. However, did not test it and it may of course triggers some undesired behaviour downstream or cause issues if Mantine changes logic. So I would rather prefer a disclaimer in dmc-docs.

AnnMarieW:
Hi @michaelw88 

Can you try running this example?  Not sure what I'm missing

```

app=Dash(external_stylesheets=dmc.styles.ALL)

data = [
    {
        "date": "Mar 22",
        "Apples": 2890,
        "Oranges": 2338,
        "Tomatoes": 2452,
    },
    {
        "date": "Mar 23",
        "Apples": 2756,
        "Oranges": 2103,
        "Tomatoes": 2402,
    },
    {
        "date": "Mar 24",
        "Apples": 3322,
        "Oranges": 986,
        "Tomatoes": 1821,
    },
    {
        "date": "Mar 25",
        "Apples": 3470,
        "Oranges": 2108,
        "Tomatoes": 2809,
    },
    {
        "date": "Mar 26",
        "Apples": 3129,
        "Oranges": 1726,
        "Tomatoes": 2290,
    },
]


component = dmc.CompositeChart(
    h=300,
    data=data,
    dataKey="date",
    withLegend=True,
    maxBarWidth=30,
    series=[
        {"name": "Tomatoes", "color": "rgba(18, 120, 255, 0.2)", "type": "bar"},
        {"name": "Apples", "color": "red.8", "type": "line"},
        {"name": "Oranges", "color": "yellow.8", "type": "area"},
    ],
    highlightHover=True
)

app.layout = dmc.MantineProvider(component)


if __name__ == "__main__":
    app.run(debug=True, port=8052)



```

michaelw88:
Not missing anything. If barProps/areaProps/lineProps were undefined, the code failed as fillOpacity will be written on undefined.

Fixed this by changing chartProps to an empty object in case props are undefined.
```
if (chartType === "bar") {
    chartProps = barProps ?? {};
} else if (chartType === "area") {
    chartProps = areaProps ?? {};
} else if (chartType === "line") {
    chartProps = lineProps ?? {};
}
```

AnnMarieW:
> Was it intentional to remove `withPointLabels`?
That's an oops.  @michaelw88 do you want me to fix?

michaelw88:
True, not intentional. Appreciated if you can fix it, being at work where
GitHub is blocked right now.

Ann Marie Ward ***@***.***> schrieb am Fr., 1. Nov. 2024,
14:08:

alexcjohnson (approved):
💃

PR #372, opened 2024-10-24
[fix] Prevent loading from empty `src` in `Avatar` component

mmarfat:
### Summary
This PR addresses an issue in the `Avatar` component where the component attempts to load a resource even when no `src` is provided. Previously, when `src` was missing or empty, it led to a `GET about:blank net::ERR_UNKNOWN_URL_SCHEME` error.

avatar

If `src` is empty or undefined, the component should not attempt to load an image. By adding this check, the component behaves more gracefully and avoids errors related to invalid URL schemes.

### Changes
- Updated the `sanitizedSrc` to return `undefined` when `src` is not provided.

### Example
```
from dash import Dash, html, _dash_renderer

app = Dash()

avatars = [
    dmc.Avatar(src="https://avatars.githubusercontent.com/u/91216500?v=4"),
    dmc.Avatar("MK"), # This shouldn't output errors into the console
    dmc.Avatar(src="https://avatars.githubusercontent.com/u/91216500?v=4", children="MK")
]

    html.Div(avatars)
)

if __name__ == '__main__':
    app.run_server(debug=True)
```

Let me know if you have any feedback or if you run into issues.

### DIFF
--- a/tests/test_avatar.py
+++ b/tests/test_avatar.py
@@ -13,4 +13,7 @@ def test_001av_avatar(dash_duo):
 
     dash_duo.start_server(app)
 
+    # Wait for the app to load
+    dash_duo.wait_for_text_to_equal("#avatar", "MK")
+
     assert dash_duo.get_logs() == []
--- a/tests/test_avatar.py
+++ b/tests/test_avatar.py
@@ -1,0 +1,15 @@
+from dash import Dash, html, _dash_renderer
+import dash_mantine_components as dmc
+
+_dash_renderer._set_react_version("18.2.0")
+
+def test_001av_avatar(dash_duo):
+    app = Dash(__name__)
+
+    avatar = dmc.Avatar("MK")
+
+    app.layout = dmc.MantineProvider(html.Div([avatar]))
+
+    dash_duo.start_server(app)
+
+    assert dash_duo.get_logs() == []
--- a/tests/test_avatar.py
+++ b/tests/test_avatar.py
@@ -6,7 +6,8 @@ _dash_renderer._set_react_version("18.2.0")
 def test_001av_avatar(dash_duo):
     app = Dash(__name__)
 
-    avatar = dmc.Avatar("MK")
+    avatar = dmc.Avatar("MK", id="avatar")
+
 
     app.layout = dmc.MantineProvider(html.Div([avatar]))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Changed
 - Reduced the Highlight component file size. #366 by @AnnMarieW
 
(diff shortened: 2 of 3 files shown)


AnnMarieW:
Hi @mmarfat 

This looks really good - Thanks!

Would you like to convert your example into a test?  You can find an example here https://github.com/snehilvj/dash-mantine-components/blob/master/tests/test_popover.py

The test for this PR would be even simpler because there are no callbacks needed.  
This line checks for errors in the console

```
 assert dash_duo.get_logs() == []
```

mmarfat:
Hi @AnnMarieW 

I've added the test but I was having some issues with `pytest` so I can't be 100% sure it passes. However, it seems simple enough so I believe it should work.

tests/test_avatar.py
```diff
@@ -0,0 +1,15 @@
+from dash import Dash, html, _dash_renderer
+import dash_mantine_components as dmc
+
+_dash_renderer._set_react_version("18.2.0")
+
+def test_001av_avatar(dash_duo):
+    app = Dash(__name__)
+
+    avatar = dmc.Avatar("MK")
```
AnnMarieW, line 9:
```suggestion
    avatar = dmc.Avatar("MK", id="avatar")

```

tests/test_avatar.py
```diff
@@ -0,0 +1,15 @@
...
+
+def test_001av_avatar(dash_duo):
+    app = Dash(__name__)
+
+    avatar = dmc.Avatar("MK")
+
+    app.layout = dmc.MantineProvider(html.Div([avatar]))
+
+    dash_duo.start_server(app)
+
```
AnnMarieW, line 14:
```suggestion

    # Wait for the app to load
    dash_duo.wait_for_text_to_equal("#avatar", "MK")

```

AnnMarieW (changes requested):
Hi @mmarfat 

This is a good start.  It just needs to make sure the component is rendered before checking for console errors.

mmarfat:
Ok, thanks. I've committed the reviewed changes.

AnnMarieW:
Great!  Looks good :slightly_smiling_face: 

Could you also please add an entry to the [CHANGELOG.md](https://github.com/snehilvj/dash-mantine-components/blob/master/CHANGELOG.md).  
Under  Unreleased, add a Fixed section and include a short description, a reference to this PR and your GitHub username.

mmarfat:
Done. I'll try to remember to do it next time without having to be reminded 😅.

AnnMarieW (approved):

alexcjohnson (approved):

AnnMarieW:
This was a nice catch - Thanks again @mmarfat :pray:

PR #379, opened 2024-10-28
Enable base64 images in Avatar

### LINKED ISSUE #374: [Feature Request] Support base 64 images in dmc.Avatar
Was trying to use dcc.Upload with dmc.Avatar but it wouldn't work, reading through the docs of dcc.Upload I noticed it uses a base 64 image that is capable of working with html.Image, attempted to use this base 64 upload in dmc.Image and it works fine, just would be a nice improvement to have dmc.Avatar also support base 64 images.

AnnMarieW:
closes #374 

Removed sanitizing url because it's not necessary for image components.

dmc.Avatar renders as an html Img

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Fixed
 - Prevent loading from empty `src` in `Avatar`. #372 by @mmarfat
+- Enable base64 images in Avatar #379 by @AnnMarieW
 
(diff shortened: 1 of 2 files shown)


alexcjohnson (approved):
Makes me a little nervous, but AFAICT yes you're right, this specific case does not pose a risk. 💃

PR #404, opened 2024-11-07
update release process

AnnMarieW:
- In release 0.14.7 I intended to [exclude the metadata.json file.](https://github.com/snehilvj/dash-mantine-components/pull/357), however, this file was still included.  In this PR I specifically excluded the file from MANIFEST.IN.  When I do `npm run dist`, this file is excluded in the  `.tar.gz`, but it's still included in the `.whl`.  I'm not sure how to fix this, but before next release, make sure the `metadata.json` file is excluded (delete the file before creating the distribution if necessary)

- I excluded the dependency "@mantine/spotlight": "7.13.4" because we don't have that component (yet), but it didn't seem to make any difference to the .js file.  Is it worth excluding?


alexcjohnson:
re `metadata.json`: per Slack, I suspect the issue is a `build` dir that should be deleted - normally right before `python setup.py sdist bdist_wheel` I run `rm -rf dist build` to prevent this.

re `@mantine/spotlight`: correct, removing it from `package.json` won't affect the bundle - the bundle only includes packages that are actually imported in the code. To be able to import a package it must be installed, therefore it must exist in `package.json`, but the converse is not true. So you can leave it there as a placeholder if you like, or you can remove it for now and put it back later, doesn't matter much.

AnnMarieW:
I just did the steps again to create the distribution files
1.  deleted the `dist` and `build` folders
2. `npm run build`
3. `npm run dist`

The wheel still has `metadata.json` included - that's what explains the difference in the size:
image

I'll update the `npm run dist` to delete the `metadata.json` first before creating the other dist files.

alexcjohnson (approved):
💃

PR #405, opened 2024-11-09
Update date components

### LINKED ISSUE #403: [Feature Request] Add new debounce prop to DatePicker
Update the DatePicker so the `debounce` prop works like in https://github.com/snehilvj/dash-mantine-components/pull/383

This is especially useful when selecting a date range.

AnnMarieW:
Closes #403 

- add `debounce=True` to `DatePicker`
- add missing `highlightToday` prop
- handle `clearable` `DateInput` when `debounce=True` 
- added test

**Notes on adding `debounce` to the `DatePicker`**

The input field in `DatePicker` behaves differently from `DateInput` in Mantine:

- In `DateInput`, users can type a date directly into the Input field (or select from a calendar).  Pressing Enter will make the input lose focus, updating the value.
- In `DatePicker` users cannot type directly into the input field. Pressing Enter or clicking on the field opens the calendar, shifting the focus to it. The value is only updated by selecting dates from the calendar.  The `n_submit` prop (counter for enter key pressed)  has limited use here and is being kept only for legacy reasons (remove in a future major release?).
- `onBlur`  Behavior:
  - In `DateInput`, the onBlur counter `n_blur` is useful for detecting when a user finishes typing or selecting a date.
  - In `DatePicker`, `n_blur` isn’t useful because the focus shifts  from the input to the calendar triggering the counter. The `n_blur` counter doesn't indicate that the user has finished entering an input, so it is not included in this component.

### DIFF
--- a/tests/dates/test_date_picker.py
+++ b/tests/dates/test_date_picker.py
@@ -1,5 +1,7 @@
+import pytest
 import time
 from selenium.webdriver.common.keys import Keys
+from selenium.common.exceptions import TimeoutException
 from dash import Dash, html,  Output, Input, _dash_renderer
 import dash_mantine_components as dmc
 
@@ -173,8 +175,8 @@ def test_002dp_datepicker(dash_duo):
     picker_elem3.send_keys(Keys.TAB, Keys.TAB, Keys.ENTER, Keys.ARROW_DOWN, Keys.ENTER, Keys.ESCAPE)
 
     # initially no change
-    time.sleep(.5)
-    dash_duo.wait_for_text_to_equal("#out-2000", "None", timeout=1)
+    with pytest.raises(TimeoutException):
+        dash_duo.wait_for_text_to_equal("#out-2000", "['2024-10-01', '2024-10-08']", timeout=1)
 
     # but do expect that it is eventually called
     dash_duo.wait_for_text_to_equal("#out-2000", "['2024-10-01', '2024-10-08']")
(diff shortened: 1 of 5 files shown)


tests/dates/test_date_picker.py
```diff
@@ -0,0 +1,183 @@
...
+
+    time.sleep(.5)
+    # select a day from the calendar with the keyboard
+    picker_elem3 = dash_duo.find_element(".mantine-Popover-dropdown")
+    picker_elem3.send_keys(Keys.TAB, Keys.TAB, Keys.ENTER)
+
+
+    # initially no change
+    time.sleep(.5)
+    dash_duo.wait_for_text_to_equal("#out-2000", "None", timeout=1)
```
alexcjohnson, line 116: Could add the nice `with pytest.raises` pattern you used in https://github.com/snehilvj/dash-mantine-components/pull/407

alexcjohnson (approved):
💃

Ferreira-B:
Great job on the update!
I may have encountered a possible bug.

I have a datePickerInput with type="range" and debouce=True. 
This datePickerInput is acting as Input of a callback. 

Now, the callback should be triggered as I click the second date on the datePickerInput (as this action also closes the date picker). The problem is that the callback is not triggered when the date picker closes, but only after I click anywhere on the page.

I don't know if this is the expected behaviour, but I thought I might report it.

AnnMarieW:
Hi @Ferreira-B 
Yes, this is expected behavior.  With `debounce=True,` the callback is triggered when the input loses focus.  With this component the input stays focused after the dropdown is closed.

For your use-case, you could try setting `debounce=False`.  The callback will trigger twice, but you could check that the date value has both a start and end date.

PR #407, opened 2024-11-10
fix debounce when deleting items

### LINKED ISSUE #406: [bug] clearing MultiSelect items with debounce
Really happy to see this feature incorporated with MultiSelect. I have one slight issue while clearing individual selections while the MultiSelect box is _not_ highlighted... this seems to ignore the debounce settings, particularly relevant for the integer value ms use case. I have a MultiSelect setup where each selection affects a series of plots, where I don't want multiple callbacks to be triggered when a user selects or deselects items too quickly as the plots are a little slow to load.

_Originally posted by @benweinberg89 in https://github.com/snehilvj/dash-mantine-components/issues/383#issuecomment-2466617520_

AnnMarieW:
Closes #406 

This PR will enable the debounce to work when clearing items when the dropdown is closed when debounce is a number of ms to delay

### DIFF
--- a/tests/combobox/test_multi_select_debounce.py
+++ b/tests/combobox/test_multi_select_debounce.py
@@ -139,3 +139,44 @@ def test_002mu_multi_select_debounce(dash_duo):
     dash_duo.wait_for_text_to_equal("#output", "Selected: []")
 
     assert dash_duo.get_logs() == []
+
+
+def test_003mu_multi_select_debounce(dash_duo):
+    app = Dash(__name__)
+
+    app.layout = dmc.MantineProvider(
+        html.Div(
+            [
+                dmc.MultiSelect(
+                    id="multi-select",
+                    data=["a", "b"],
+                    value=["a"],
+                    debounce=2000
+                ),
+                html.Div(id="output"),
+            ]
+        )
+    )
+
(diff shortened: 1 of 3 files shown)


tests/combobox/test_multi_select_debounce.py
```diff
@@ -139,3 +139,44 @@ def update_output(selected_values):
...
+    # Wait for the app to load
+    dash_duo.wait_for_text_to_equal("#output", "Selected: ['a']")
+
+    # delete selected item by clicking on the pill
+    button = dash_duo.find_element("button.mantine-Pill-remove")
+    button.click()
+
+    # Verify the value updates after debounce time when the input is not focused
+    with pytest.raises(TimeoutException):
+        dash_duo.wait_for_text_to_equal("#output", '["a"]', timeout=1)
```
alexcjohnson, line 176:
```suggestion
        dash_duo.wait_for_text_to_equal("#output", "Selected: []", timeout=1)
```
I think this is what you mean, ie the same test as below just time-limited to before the debounce fires?

AnnMarieW: That would fail, because during the debounce the inital value doesn't change.  So need to design the test to show the debounce is working

PR #412, opened 2024-11-13
Enable data and value to be updated in same callback

### LINKED ISSUE #411: Can't update dmc.Select 'data' and 'value' inside the same callback
Hi,

Since 0.14.7, when using dmc.Select component in a callback, you can't output both 'data' and 'value' : value doesn't get updated.

Sample to reproduce bug : 

```
from dash import Dash, Input, Output, callback, _dash_renderer

app = Dash(__name__)

    dmc.Center(
        [
            dmc.Select(
                id="select",
                data = [],
                clearable=True, 
                nothingFoundMessage="Nothing Yet"
            ),
            dmc.Button("Update Select", id="update-select"),
        ],
        style={"height": '100vh', "width": "100vw"},
    )
)


@callback(
    Output("select", "data"),
    Output("select", "value"),
    Input("update-select", "n_clicks"),
    prevent_initial_call=True
)
def update_select(n_clicks):

    new_options = [
        {"value": "a", "label": "Option A"},
        {"value": "b", "label": "Option B"},
        {"value": "c", "label": "Option C"},
    ]
    new_value = "b"
    return new_options, new_value

if __name__ == "__main__":
    app.run_server(debug=True)
```

Spliting the update into 2 callbacks is a workaround : 

```
@callback(
    Output("select", "data"),
    Input("update-select", "n_clicks"),
    prevent_initial_call=True
)
def update_data(n_clicks):

    new_options = [
        {"value": "a", "label": "Option A"},
        {"value": "b", "label": "Option B"},
        {"value": "c", "label": "Option C"},
    ]
    
    return new_options

@callback(
    Output("select", "value"),
    Input("select", "data"),
    prevent_initial_call=True
)
def update_value(data):

    return "b"
```

I have noticed the value gets updated at first, but then something replace it by None : 

```
@callback(
    Output("select", "data"),
    Output("select", "value"),
    Input("update-select", "n_clicks"),
    prevent_initial_call=True
)
def update_select(n_clicks):

    new_options = [
        {"value": "a", "label": "Option A"},
        {"value": "b", "label": "Option B"},
        {"value": "c", "label": "Option C"},
    ]
    new_value = "b"
    return new_options, new_value

@callback(
    Input("select", "value")
)
def update_select(val):

    print(val)
```
> b
> None

AnnMarieW:
Closes #411 

In 0.14.7, there was a regression where it was not possible to update both the value and the data in the same callback in `Select` and `MultiSelect`

This issue occurred because there are two calls to `setSelected`, and in PR #383, the order in which these calls were made changed.  I considered combining them into a single `useDidUpdate` function, but that approach would have led to excessive calls to [this function](https://github.com/snehilvj/dash-mantine-components/blob/master/src/ts/components/core/combobox/Select.tsx#L98) for clearing invalid selected values when data is updated in a callback.

Todo
- [x] tests
- [x] changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 -  In MultiSelect, enable the debounce to work when deleting items when the dropdown is closed when debounce is a number. #407 by @AnnMarieW
-
+-  In MultiSelect and Select, fixed regression where it was not possible to update both the value and data in the same callback #412
 
 # 0.14.7
 
--- a/src/ts/components/core/combobox/MultiSelect.tsx
+++ b/src/ts/components/core/combobox/MultiSelect.tsx
@@ -84,10 +84,6 @@ const MultiSelect = (props: Props) => {
     }, [debounced]);
 
 
-    useDidUpdate(() => {
-            setSelected(value ?? []);
-        }, [value]);
-
     const handleKeyDown = (ev) => {
         if (ev.key === "Enter") {
             setProps({
@@ -109,6 +105,10 @@ const MultiSelect = (props: Props) => {
         const filteredSelected = filterSelected(data, selected);
         setSelected(filteredSelected ?? []);
     }, [data]);
+
+    useDidUpdate(() => {
+        setSelected(value ?? []);
+    }, [value]);
 
     useDidUpdate(() => {
         setProps({ data: options });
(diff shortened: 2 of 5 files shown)


alexcjohnson (approved):
💃 This looks good to me. If you wanted to be extra cautious, you could add a counter to the callbacks in the tests (like in the dash test [cbsc001](https://github.com/plotly/dash/blob/95665785f184aba4ce462637c30ccb7789280911/tests/integration/callbacks/test_basic_callback.py#L31) for example, to verify that the two `useDidUpdate` calls don't result in two callback calls. But I'm confident that's not going to be an issue so feel free to ignore this suggestion 😁

Issue #410, opened 2024-11-12
Potential breaking changes


AnnMarieW:
In line with our roadmap, we’re committed to minimizing breaking changes. However, I wanted to open a discussion about potential breaking changes that may come up.

### DatePicker Component
Currently, the `DatePicker` in DMC corresponds to the [`DatePickerInput`](https://mantine.dev/dates/date-picker-input/) component from Mantine. Mantine also includes a stand-alone [DatePicker](https://mantine.dev/dates/date-picker/) that functions solely as a calendar, separate from the input dropdown. There was a request and a strong use case shared [on Discord for adding this stand-alone DatePicker.](https://discord.com/channels/932931236882677780/1305692070907875338)

To avoid a breaking change, should we consider naming the new stand-alone `DatePicker` something else? On the other hand, to keep DMC aligned with the upstream Mantine library, it might be better to rename the existing `DatePicker` to `DatePickerInput`.

demk0r:
The standalone DatePicker does have valid uses cases, so it would be great to add it to DMC. I recently also stumbled over this and was confused why the naming is not consistent between DMC and Mantine. I would prefer a breaking change in DMC to keep alignment with Mantine, since the necessary code adjustment for users would be a simple search and replace.

AnnMarieW:
**Update**
By popular demand, the `DatePicker` will be renamed to `DatePickerInput` in the next release. 0.15.0.  For more info see [Release Notes](https://github.com/snehilvj/dash-mantine-components/discussions/415)

PR #414, opened 2024-11-14
DatePicker renamed to DatePickerInput

AnnMarieW:
This PR renames the DatePicker to DatePickerInput so the component name is aligned with the upstream Mantine library.
For more information see #410

Until the real DatePicker component is added, you will see an error message:

image

Here is a sample app to show the error message for the old component name, and that the new component works as expected
```python
from datetime import datetime

component = dmc.Stack(
    [
        dmc.DatePicker(
            value=datetime.now().date(),
            label="Date not clearable",
           # w=200,
        ),
        dmc.DatePickerInput(
            value=datetime.now().date(),
            label="Date clearable",
            w=200,
            clearable=True,
        ),
    ]
)


    component
)

if __name__ == "__main__":
    app.run(debug=True)
    
 ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,19 @@
 
 # Unreleased
 
+### Added
+
 - New additons to Date components #405 by @AnnMarieW
   - `DatePicker` - The `debounce` prop can now be `True` `False` or `number` of ms delay before updating.  When True, the value updates on blur. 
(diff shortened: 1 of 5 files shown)


AnnMarieW:
Docs PR here:  https://github.com/snehilvj/dmc-docs/pull/106

alexcjohnson (approved):
💃 Looks good to me!

PR #416, opened 2024-11-15
update to mantine 7.14.1

AnnMarieW:
This updates Mantie from 7.13.3 to 7.14.1
- added `restrictToMarks` prop to `Slider`  (Note this prop is not yet available for RangeSlider)
- added `barLabelColor` to `BarChart`

I think the next release should be 0.15.0 given that it's a minor release from Mantine, plus there is a a breaking change.


AnnMarieW:
- `restrictToMarks` prop for Slider:

Here is a sample app on PyCafe:  https://py.cafe/amward/dmc-slider-with-restrict-to-marks

alexcjohnson (approved):
💃

PR #419, opened 2024-11-17
Add position prop to Notification

AnnMarieW:
This prop was added in V7.12:  https://mantine.dev/changelog/7-12-0/

Here's a sample app for the docs:

See it live:  https://py.cafe/amward/dmc-PR-419-Notification-Position

```python
from dash import Dash, html, callback, Output, Input, _dash_renderer

positions = ['top-left', 'top-right', 'bottom-left', 'bottom-right', 'top-center', 'bottom-center']

app.layout = dmc.MantineProvider(html.Div(
    [
        dmc.NotificationProvider(withinPortal=True),
        dmc.Center(
            dmc.Select(label="Notification Position", data=positions, value="top-right", id="notify-position", w=400),
            style={"height": 300, "width": "100%"},
        ),
        html.Div(id="notify-container")
    ]
))

@callback(
    Output("notify-container", "children"),
    Input("notify-position", "value"),
    prevent_initial_call=True,
)
def notify(value):
    return dmc.Notification(
            title=f"Notification {value}",
            autoClose=False,
            action="show",
            message="Hello World",
            position=value
        )


if __name__ == "__main__":
    app.run(debug=True)


```


alexcjohnson (approved):
💃

PR #431, opened 2024-11-23
add mantine_light and mantine_dark plotly figure templates

AnnMarieW:
This PR adds the `create_mantine_figure_templates()` function which creates Mantine-styled Plotly figure templates for both light and dark modes using the default Mantine theme.  It registers the templates with plotly.io.templates as "mantine_light" and "mantine_dark", and optionally sets one of the themes as a default.

The templates can be used in an app just like any built-in plotly template, for example:

```
px.histogram(dff, x='lifeExp', title='2007 Distribution of Life Expectancy', template="mantine_dark")
```

Here is a sample app hosted on PyCafe: https://py.cafe/amward/mantine-themed-plotly-figure-templates2

image

image

### DIFF
--- a/dash_mantine_components/figure_templates.py
+++ b/dash_mantine_components/figure_templates.py
@@ -4,7 +4,7 @@ import plotly.io as pio
 import copy
 
 
-def create_mantine_figure_templates(default=None):
+def add_figure_templates(default=None):
 
     """
     Create and register Plotly figure templates styled to match the Mantine default theme.
--- a/dash_mantine_components/figure_templates.py
+++ b/dash_mantine_components/figure_templates.py
@@ -1,0 +1,94 @@
+import dash_mantine_components as dmc
+import plotly.graph_objects as go
+import plotly.io as pio
+import copy
+
+
+def create_mantine_figure_templates():
+    """
+    Create Mantine-styled Plotly templates for both light and dark modes.
+    registers templates with plotly.io.templates as "mantine_light" and "mantine_dark"
+    """
+
+    colors = dmc.DEFAULT_THEME["colors"]
+    font_family = dmc.DEFAULT_THEME["fontFamily"]
+    # pallet generated from https://www.learnui.design/tools/data-color-picker.html#palette
+    custom_colorscale = [
+        "#1864ab",  # blue[9]
+        "#7065b9",
+        "#af61b7",
+        "#e35ea5",
+        "#ff6587",
+        "#ff7c63",
+        "#ff9e3d",
+        "#fcc419",  # yellow[5]
+    ]
+
+    # Default theme configurations
+    default_themes = {
+        "light": {
+            "colorway": [
+                colors[color][6]
+                for color in ["blue", "red", "green", "violet", "orange", "cyan", "pink", "yellow"]
+            ],
+            "paper_bgcolor": "#f8f9fa",
+            "plot_bgcolor": "#ffffff",
+            "gridcolor": "#dee2e6",
+        },
+        "dark": {
+            "colorway": [
+                colors[color][8]
+                for color in ["blue", "red", "green", "violet", "orange", "cyan", "pink", "yellow"]
+            ],
+            "paper_bgcolor": "#1a1b1e",
+            "plot_bgcolor": "#25262b",
+            "gridcolor": "#343a40",
+        }
+    }
+
+    def make_template(name):
+        #Start with either a light or dark Plotly template
+        base = "plotly_white" if name == "light" else "plotly_dark"
+        template = copy.deepcopy(pio.templates[base])
+
+        layout = template.layout
+        theme_config = default_themes[name]
+
+        # Apply theme settings
+        layout.colorway = theme_config["colorway"]
+        layout.colorscale.sequential = custom_colorscale
+        layout.piecolorway = theme_config["colorway"]
+        layout.paper_bgcolor = theme_config["paper_bgcolor"]
+        layout.plot_bgcolor = theme_config["plot_bgcolor"]
+        layout.font.family = font_family
+
+        # Grid settings
+        for axis in (layout.xaxis, layout.yaxis):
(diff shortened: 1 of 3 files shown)


AnnMarieW:
Here's a demo app for the docs:  https://py.cafe/amward/mantine-plotly-figure-templates

dash_mantine_components/figure_templates.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # #register templates
+    pio.templates["mantine_light"] = make_template("light")
+    pio.templates["mantine_dark"] = make_template("dark")
+
+    return None
+
+
+# Create and register templates
+create_mantine_figure_templates()
```
alexcjohnson, line 94: I'm a little uneasy about automatically calling this - the convenience is great, but it's a bit magical. For example if you like these themes and want to use them without DMC, you have to know to import the package anyway and silence your linter.

Curious if perhaps @ndrezn you know whether any other packages add their own templates like this, or if you have thoughts on this pattern?

If we do automatically call this, there's no reason to import it by name in `__init__.py` - just `import .figure_templates` will do.

dash_mantine_components/figure_templates.py
```diff
@@ -0,0 +1,94 @@
+import dash_mantine_components as dmc
```
alexcjohnson, line 1: Apparently this works, but it's a circular reference. Better to get what you need directly from `.theme`, like `from .theme import DEFAULT_THEME` as it is in `__init__.py`

dash_mantine_components/figure_templates.py
AnnMarieW: I started with requiring it to be imported in an app, but thought it was an unnecessary line of code when it could just be made available, but I see your point about it being too magical.

I can't imagine why anyone would use this outside of DMC though.  It's not different enough from `plotly_white` and `plotly_dark` themes.  It's just tweaked a little to make it look better with the color tones used in the default DMC theme.

ndrezn: Maybe a flag in `MantineProvider` like `style_charts` would be better?

AnnMarieW: If it's documented that when you use the dmc library, you will have access to the "mantine_light" and "mantine_dark" templates, shouldn't that be OK?, Why is that  any more magical than having access to all  the other built-in templates that come with the Plotly library?   They are not set as defaults -  you have to specify them in the figure definition otherwise you get the standard "plotly" template.

alexcjohnson: It feels somewhat magical to me because with the built-in templates you can go look for docs on the `template` arg and you immediately get to https://plotly.com/python/templates/#view-available-themes - which lists all the built-in templates, and you can go from there to learn about each of these templates. With these ones, an operation that normally has no side-effects (importing a library) results in extra options being available in a different library, so if you want to learn about these or if they disappeared and you want to know how to get them back, what do you do?

That said, given that the new options have the name `mantine` in them and that's a pretty distinctive name, it probably wouldn't be too hard for users to connect the dots and figure out that they came from `dash_mantine_components` - then if it's obvious how to find them in the docs we're good. Also we document exactly the pattern you're using 😅 https://plotly.com/python/templates/#saving-and-distributing-custom-themes

So OK, let's leave it as is. Then I'd still change the import in `__init__.py` to just `import .figure_templates` 😎 

> a reduced import if you want to use just the named themes but not the rest of Mantine

I don't think we can do this - importing a submodule will [always import the parent](https://docs.python.org/3/reference/import.html#:~:text=Importing%20parent.one%20will%20implicitly%20execute%20parent/__init__.py%20and%20parent/one/__init__.py.%20Subsequent%20imports%20of%20parent.two%20or%20parent.three%20will%20execute%20parent/two/__init__.py%20and%20parent/three/__init__.py%20respectively.) - so unless we wanted to move the rest of DMC out of the top-level module we're always going to get it. Anyway I guess @AnnMarieW is right that this is similar enough to built-in themes that this is unlikely to come up much.

AnnMarieW: ah, OK, I get it now:
> With these ones, an operation that normally has no side-effects (importing a library) results in extra options being available in a different library,

You're right - it's better to make this a function that is called in the app.   Also, if it's a function, I could make it so you could set the default to one of these templates, which would be very convenient when not using a theme switcher.

Next trick would be to make it possible to redraw the figure when the theme changes without updating the entire figure in a callback, but I think that would need a custom `Graph` component based on `dcc.Graph`

dash_mantine_components/figure_templates.py
```diff
@@ -0,0 +1,110 @@
...
+        return template
+
+
+    # #register templates
+    pio.templates["mantine_light"] = make_template("light")
+    pio.templates["mantine_dark"] = make_template("dark")
+
+    # set the default
+    if default in ["mantine_light", "mantine_dark"]:
+        pio.templates.default = default
```
alexcjohnson, line 108:
```suggestion
    if default in ["mantine_light", "mantine_dark"]:
        pio.templates.default = default
    elif default:
        raise ValueError(f"unrecognized {default=}, allowed values are 'mantine_light' and 'mantine_dark'")
```
(I just learned about the `=` [format option](https://pythonprinciples.com/ask/f-strings-what-does-an-equals-sign-do/) the other day, it's my new favorite!)

AnnMarieW: Ooh nice tip - thanks!

dash_mantine_components/figure_templates.py
AnnMarieW: I agree, I wasn’t a fan of the name, but couldn’t think of a better one.  I like your suggestion 🙂

AnnMarieW:
@alexcjohnson Do you have any comments on the style of the new figure templates?  Anything that should be  changed?  Anything missing for other chart types?

PR #441, opened 2024-12-03
enable searchValue to be updated in callbacks

AnnMarieW:
This enables a [feature request posted on the forum](https://community.plotly.com/t/issue-with-search-text-reset-in-dash-mantine-components-multiselect/89133) to make it possible to keep the search value after an item is selected  in MultiSelect.

The dmc.MultiSelect automatically clears the search value when an item is selected, but it should be possible to restore it in a callback.  

This PR fixes a bug where it was not possible to update the `searchValue` in a callback.

Here is a sample use case:

https://py.cafe/amward/dmc-pr-441-update-searchValue-in-callback

```python

from dash import Dash, _dash_renderer, html, dcc, Input, Output, State, callback

component = dmc.MultiSelect(
    label="Pick your favorite libraries",
    data=["Pandas", "NumPy", "TensorFlow", "PyTorch"],
    searchable=True,
    w=400,
    id="select",
)


    dcc.Store(id="search"),
    html.Div(id="out"),
    component
])

@callback(
    Output("search", "data"),
    Input("select", "searchValue")
)
def update(searchValue):
    return searchValue


@callback(
    Output("out", "children"),
    Output("select", "searchValue"),
    Input("select", "value"),
    State("search", "data")
)
def update(val,search):
    return f"You selected {val}", search

```

### DIFF
--- a/src/ts/components/core/combobox/MultiSelect.tsx
+++ b/src/ts/components/core/combobox/MultiSelect.tsx
@@ -58,14 +58,12 @@ const MultiSelect = (props: Props) => {
         n_submit,
         n_blur,
         data,
-        searchValue,
         value,
         ...others
     } = props;
 
     const [selected, setSelected] = useState(value);
     const [options, setOptions] = useState(data);
-    const [searchVal, setSearchVal] = useState(searchValue);
     const { ref, focused } = useFocusWithin();
 
     const debounceValue = typeof debounce === "number" ? debounce : 0;
@@ -82,12 +80,6 @@ const MultiSelect = (props: Props) => {
             setProps({ value: debounced });
         }
     }, [debounced]);
-
-    useDidUpdate(() => {
-        if (searchValue !== searchVal) {
-            setSearchVal(searchValue);
-        }
-    }, [searchValue]);
 
     const handleKeyDown = (ev) => {
         if (ev.key === "Enter") {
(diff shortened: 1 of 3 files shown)


src/ts/components/core/combobox/MultiSelect.tsx
```diff
@@ -96,10 +101,15 @@ const MultiSelect = (props: Props) => {
...
         setProps({
             n_blur: n_blur + 1,
-            ...(debounce === true && { value: selected })
+            ...(debounce === true && { value: selected }),
         });
     };
 
+    const handleSearchChange = (newSearchVal) => {
+        setSearchVal(newSearchVal);
+        setProps({ searchValue: newSearchVal });
```
alexcjohnson, line 110: If you're setting both of these simultaneously now, do we even need to have `searchValue` copied into state? Can't we just use the prop directly?

AnnMarieW: I deleted the `setSearchVal(newSearchVal);` and it works :-)

alexcjohnson: I was making a stronger statement actually - that maybe now that `handleSearchChange` is updating the prop we don't need
```py
const [searchVal, setSearchVal] = useState(searchValue);
```
or the rest of the state management it entails at all, and we could just forward `searchValue` directly to the component as `searchValue={searchValue}` or even leave it in `others`.

AnnMarieW: oh, cool!  That simplifies things a lot!

alexcjohnson (approved):
💃

PR #445, opened 2024-12-05
added ColorSchemeSwitch

AnnMarieW:
A toggle switch component for changing between light and dark color schemes.
  
 The `ColorSchemeSwitch` allows users to toggle between light and dark modes
  with a single click. No callbacks required!  It uses icons to indicate the available action: the icon
  displayed represents the mode the user can switch to (e.g., a moon icon for
  toggling to dark mode and a sun icon for toggling to light mode).
  
 Behavior:
  - Automatically syncs with the application's color scheme using Mantine's context.
  - Updates the color scheme via [`setColorScheme`](https://mantine.dev/theming/color-schemes/) when toggled or when `color_scheme` changes.
     -- The color scheme selected by the is persistent     
  - Displays appropriate icons for toggling actions

```

component = dmc.ColorSchemeSwitch(
    iconForLightMode=DashIconify(icon="radix-icons:moon", width=25),
    iconForDarkMode=DashIconify( icon="radix-icons:sun",width=25),
    color="yellow",
)

    dmc.Group([
        dmc.Title("ColorSchemeSwitch Demo"),
        component
    ], p="lg"),
)

if __name__ == "__main__":
    app.run(debug=True)


```

dmc-ColorSchemeSwitch

### DIFF
--- a/src/ts/components/styles/ColorSchemeSwitch.tsx
+++ b/src/ts/components/styles/ColorSchemeSwitch.tsx
@@ -1,0 +1,78 @@
+import {
+    useComputedColorScheme,
+    useMantineColorScheme,
+    ActionIcon as MantineActionIcon,
+} from "@mantine/core";
+import { ActionIconProps } from "props/actionicon";
+import { DashBaseProps } from "props/dash";
+import React, { useEffect } from "react";
+
+interface Props extends ActionIconProps, DashBaseProps {
+    /** An integer that represents the number of times that this element has been clicked on */
+    n_clicks?: number;
+    /** Current color schem  "light" | "dark" */
+    color_scheme?: "light" | "dark";
+    /** The icon to display when the application is in light mode. This icon represents the action to switch to dark mode.  Default: moon icon */
+    iconForLightMode: React.ReactNode;
+    /** The icon to display when the application is in dark mode. This icon represents the action to switch to light mode.  Default: sun icon */
+    iconForDarkMode: React.ReactNode;
+}
+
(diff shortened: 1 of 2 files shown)


AnnMarieW:
@alexcjohnson 
I have some questions about this component's API before I write some tests...

I'm struggling with the prop names for  icons.  For this `ActionIcon`, in dark mode, you show something like a sun to indicate to click on it to switch to light mode.   So should an icon like the sun be called a `lightIcon` or should it be called the `iconForDarkMode` ?

Also, I'd like to enable the use of a `Switch` like this one with [icon labels](https://mantine.dev/core/switch/#icon-labels).   Should this be a separate component?  Or should it be a prop, something like `component="switch"` or `component="actionicon"`

alexcjohnson:
I think I'd find it confusing to have a light/dark switch that's just a single icon. The switch you show is better but clearer still IMO is what we have in the main Dash docs, a toggle switch with an icon on either side. That also removes all the confusion about what to call the icons.

dash-light-dark

Even if you do want to keep the single-icon view, I'd name the icons based on what's in them, so the sun is `lightIcon` and the moon is `darkIcon`.

AnnMarieW:
I'm just replicated what they have on the Mantine docs site.
I checked the [Bootstrap docs](https://getbootstrap.com/), they have a single icon too (with a dropdown)  but the icon indicates the current state of the color mode.  :thinking: 

Yeah, I think this is too confusing and it will be better to use  this Mantine switch https://mantine.dev/core/switch/#icon-labels
image
image

Issue #390, opened 2024-11-02
Performance issues


CNFeffery:
I recently noticed some potential performance issues, which involve the `MantineProvider` of `dmc`. Related issue: https://github.com/plotly/dash/issues/3057

CNFeffery:
for example. everything will be re-rendered when I just click the top-left button:
![image](https://github.com/user-attachments/assets/35dd923a-8978-4026-a297-d7073e5a3d09)

AnnMarieW:
Hi @CNFeffery 
Thanks for reporting... do you know if there is anything we can do here to fix this, or is it an upstream issue?

AnnMarieW:
Hey @Spriteware   Since you did such an amazing job with performance improvements for Dash pattern matching callbacks -- do you have any thoughts about this one?

CNFeffery:
> Hi @CNFeffery Thanks for reporting... do you know if there is anything we can do here to fix this, or is it an upstream issue?

I believe this is an issue that can only be addressed from the underlying level of dash.

AnnMarieW:
Great news!  Fix is in progress :tada: 
https://github.com/plotly/dash/pull/3066

AnnMarieW:
This is fixed in dash 3.0

PR #458, opened 2024-12-12
Update for Dash 3.0:  Remove dashprivate props, change loading state

### LINKED ISSUE #453: Update for Dash 3.0
We are using `dashprivate` props is several places.  This should be replaced with context components because `dashprivate` may be removed at any time.  More info in [Dash pr 3066](https://github.com/plotly/dash/pull/3066#issuecomment-2529099478)

Here's where we are currently using [context](https://github.com/search?q=repo%3Asnehilvj%2Fdash-mantine-components%20context&type=code)

Here's where we are currently using [dashprivate](https://github.com/search?q=repo%3Asnehilvj%2Fdash-mantine-components+dashprivate&type=code)

Might want to wait on this one until the Plotly team has a recommended solution.  There may be some new dash functions that will be easier than using context.

AnnMarieW:
### Note:  This PR will be closed in favor of #499

Close #453 

When updates were made in [Dash PR#3066  ](https://github.com/plotly/dash/pull/3066) to address  #390, `_dashprivate` props were removed.  This PR updates to remove these props from dmc.  Also loading state coming from the dash render has changed.

- [x] Checkbox - update function to handle custom icons in checkboxes
- [x] Popover
- [x] Menu
- [x] Hovercard
- [x] Timeline
- [x] Stepper
- [x]  Anchor - This component uses `_dashprivate_pushstate`.  Confirmed that it's OK to use this private variable.  It enables the `dmc.Anchor` to work like the `dcc.Link` component with the `dcc.Location` component.  See [3066 comment](https://github.com/plotly/dash/pull/3066#issuecomment-2546077821)
- [ ] Skeleton.  This component works when you manually set the `visible` prop, but does not automatically update based on the loading status of it's children.  Would probably need to add logic similar to the dcc.Loading component.
- [x] Check on the use of [`renderDashComponents` ](https://github.com/emilhe/dash-extensions-js/blob/8b07ce3b61bbd3150cbf8dd6c5fb6d0204e74adf/index.js#L372) function and see if there is a better way.  It's used in these  [4 components](https://github.com/search?q=repo%3Asnehilvj%2Fdash-mantine-components+renderDash&type=code) .  See [this comment](https://github.com/plotly/dash/pull/3066#issuecomment-2545768188) in Dash PR.  Chatted with Philippe and he mentioned that he could add `render(component, path)` to `dash_component_api` quite easily, but it would be in a release after Dash 3.0.  I verified that the 4 components using `renderDashComponents` still works with Dash PR 3066.

- [x] Dash 3 uses React 18.3.1 which no longer supports `defaultProps`.  Need to update every component and test the docs app with React 18.3.1.  Updated all but `Drawer` and `Modal`
- [ ] Can't set defaults on required props without `defaultProps`.  Affects `Drawer`, `Modal`,

- [ ] Update new components after they are merged:  `Tree`  `InlineCodeHighlight`
- [ ] See [this comment](https://github.com/plotly/dash/pull/3066#discussion_r1928923549) about setting default props for persistence 
- [ ] Update `get_layout`  See [this comment](https://github.com/plotly/dash/pull/3066#issuecomment-2612884559)

For setting defaults on required types in Mantine try:
> set it to Partial<MantineProps> and all the props would be optional, if only want some types can take those props and the rest as is.

Note - Test pass locally when using the dash version from PR 3066

### DIFF
--- a/src/ts/components/charts/AreaChart.tsx
+++ b/src/ts/components/charts/AreaChart.tsx
@@ -65,7 +65,7 @@ interface Props
 
 /** AreaChart */
 const AreaChart = (props: Props) => {
-    const { setProps, loading_state, clickData, hoverData, clickSeriesName, hoverSeriesName, series, highlightHover, areaChartProps, activeDotProps, areaProps, ...others } = props;
+    const { setProps, clickData, hoverData, clickSeriesName, hoverSeriesName, series, highlightHover, areaChartProps, activeDotProps, areaProps, ...others } = props;
 
     const [highlightedArea, setHighlightedArea] = useState(null);
     const shouldHighlight = highlightHover && highlightedArea !== null;
@@ -151,9 +151,12 @@ const AreaChart = (props: Props) => {
 
     const newProps = { ...areaChartProps, onClick, onMouseOver };
 
+    const ctx = (window as any).dash_component_api.useDashContext();
+    const loading = ctx.useLoading();
+
     return (
       <MantineAreaChart
-        data-dash-is-loading={(loading_state && loading_state.is_loading) || undefined}
+        data-dash-is-loading={loading || undefined}
         areaChartProps={newProps}
         series={series}
         activeDotProps={{
--- a/src/ts/components/charts/AreaChart.tsx
+++ b/src/ts/components/charts/AreaChart.tsx
@@ -65,7 +65,7 @@ interface Props
 
 /** AreaChart */
 const AreaChart = (props: Props) => {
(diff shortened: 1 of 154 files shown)


AnnMarieW:
Checkbox test app
```python
from dash import Dash, _dash_renderer, Input, Output, State

component = dmc.Stack([
    dmc.Checkbox(
        label="Custom checked icon",
        checked=True,
        icon=DashIconify(icon="ion:bag-check-sharp"),
        size="lg",
        p=0,
    ),
    dmc.Checkbox(
        label="Custom intermediate icons",
        indeterminate=True,
        indeterminateIcon=DashIconify(icon="material-symbols:indeterminate-question-box"),
        size="lg"
    )
])

    component
)

if __name__ == "__main__":
    app.run(debug=True)

```

PR #460, opened 2024-12-14
Add tree component

Godisemo:
Add tree component.

Example app below
```python
from dash import Input, Output, callback, Dash, _dash_renderer
import json

app = Dash()

data = [
    {
        "label": "src",
        "value": "src",
        "children": [
            {
                "label": "components",
                "value": "src/components",
                "children": [
                    {"label": "Accordion.tsx", "value": "src/components/Accordion.tsx"},
                    {"label": "Tree.tsx", "value": "src/components/Tree.tsx"},
                    {"label": "Button.tsx", "value": "src/components/Button.tsx"},
                ],
            },
        ],
    },
    {
        "label": "node_modules",
        "value": "node_modules",
        "children": [
            {
                "label": "react",
                "value": "node_modules/react",
                "children": [
                    {"label": "index.d.ts", "value": "node_modules/react/index.d.ts"},
                    {
                        "label": "package.json",
                        "value": "node_modules/react/package.json",
                    },
                ],
            },
            {
                "label": "@mantine",
                "value": "node_modules/@mantine",
                "children": [
                    {
                        "label": "core",
                        "value": "node_modules/@mantine/core",
                        "children": [
                            {
                                "label": "index.d.ts",
                                "value": "node_modules/@mantine/core/index.d.ts",
                            },
                            {
                                "label": "package.json",
                                "value": "node_modules/@mantine/core/package.json",
                            },
                        ],
                    },
                    {
                        "label": "hooks",
                        "value": "node_modules/@mantine/hooks",
                        "children": [
                            {
                                "label": "index.d.ts",
                                "value": "node_modules/@mantine/hooks/index.d.ts",
                            },
                            {
                                "label": "package.json",
                                "value": "node_modules/@mantine/hooks/package.json",
                            },
                        ],
                    },
                    {
                        "label": "form",
                        "value": "node_modules/@mantine/form",
                        "children": [
                            {
                                "label": "index.d.ts",
                                "value": "node_modules/@mantine/form/index.d.ts",
                            },
                            {
                                "label": "package.json",
                                "value": "node_modules/@mantine/form/package.json",
                            },
                        ],
                    },
                ],
            },
        ],
    },
    {
        "label": "package.json",
        "value": "package.json",
    },
    {
        "label": "tsconfig.json",
        "value": "tsconfig.json",
    },
]

    [
        dmc.Group(
            [
                dmc.Tree(
                    id="tree",
                    allowRangeSelection=False,
                    # checkOnSpace=True,
                    clearSelectionOnOutsideClick=True,
                    # expandOnClick=False,
                    # expandOnSpace=False,
                    levelOffset="xl",
                    selectOnClick=True,
                    data=data,
                    checkboxes=True,
                    checked=["node_modules/@mantine/form/index.d.ts"],
                    selected=["node_modules/@mantine/form/index.d.ts"],
                    expanded=[
                        "node_modules",
                        "node_modules/@mantine",
                        "node_modules/@mantine/form",
                        "node_modules/@mantine/form/index.d.ts",
                    ],
                ),
                dmc.CodeHighlight(id="expanded", code="", language="json"),
                dmc.CodeHighlight(id="checked", code="", language="json"),
                dmc.CodeHighlight(id="selected", code="", language="json"),
            ],
            align="flex-start",
            grow=True,
        ),
    ],
    theme={"primaryColor": "pink"},
)


@callback(
    Output("expanded", "code"),
    Output("checked", "code"),
    Output("selected", "code"),
    Input("tree", "expanded"),
    Input("tree", "checked"),
    Input("tree", "selected"),
)
def tree_callback(expanded, checked, selected):
    expanded_code = json.dumps({"expanded": expanded}, indent=4)
    checked_code = json.dumps({"checked": checked}, indent=4)
    selected_code = json.dumps({"selected": selected}, indent=4)
    return expanded_code, checked_code, selected_code


if __name__ == "__main__":
    app.run(debug=True)


```

### DIFF
--- a/src/ts/components/core/Tree.tsx
+++ b/src/ts/components/core/Tree.tsx
@@ -15,11 +15,30 @@ import { StylesApiProps } from "props/styles";
 import React from "react";
 
 interface Props extends BoxProps, StylesApiProps, DashBaseProps {
-    data: TreeNodeData[];
-    expanded?: string[] | "*";
-    checked?: string[];
+    /** Determines whether tree nodes range can be selected with click when Shift key is pressed, `true` by default */
+    allowRangeSelection?: boolean;
+    /** Determines if checkboxes should be rendered, `false` by default */
     checkboxes?: boolean;
+    /** Determines checked nodes as a list of values (note that only leaves can be checked), `[]` by default */
+    checked?: string[];
+    /** Determines whether tree node should be checked on space key press, `false` by default */
+    checkOnSpace?: boolean;
+    /** Determines whether selection should be cleared when user clicks outside of the tree, `false` by default */
+    clearSelectionOnOutsideClick?: boolean;
+    /** Data used to render nodes */
+    data: TreeNodeData[];
+    /** Determines expanded nodes as a list of values or `'*'` for all, `[]` by default */
+    expanded?: string[] | "*";
+    /** Determines whether tree node with children should be expanded on click, `true` by default */
+    expandOnClick?: boolean;
+    /** Determines whether tree node with children should be expanded on space key press, `true` by default */
+    expandOnSpace?: boolean;
+    /** Horizontal padding of each subtree level, key of `theme.spacing` or any valid CSS value, `'lg'` by default */
     levelOffset?: string | number;
+    /** Determines selected nodes as a list of values, `[]` by default */
+    selected?: string[];
+    /** Determines whether node should be selected on click, `false` by default */
+    selectOnClick?: boolean;
 }
 
 const Leaf = ({
@@ -29,7 +48,8 @@ const Leaf = ({
     elementProps,
     tree,
     checkboxes,
-}: RenderTreeNodePayload & { checkboxes: boolean }) => {
+    expandOnClick,
+}: RenderTreeNodePayload & { checkboxes: boolean; expandOnClick: boolean }) => {
     const checked = tree.isNodeChecked(node.value);
     const indeterminate = tree.isNodeIndeterminate(node.value);
     return (
@@ -45,7 +65,14 @@ const Leaf = ({
                     }
                 />
             )}
-            <Group gap={5} onClick={() => tree.toggleExpanded(node.value)}>
+            <Group
+                gap={5}
+                onClick={
+                    expandOnClick
+                        ? () => tree.toggleExpanded(node.value)
+                        : undefined
+                }
+            >
                 <span>{node.label}</span>
                 {hasChildren && (
                     <AccordionChevron
@@ -67,7 +94,9 @@ const Tree = (props: Props) => {
         checked,
         data,
         expanded,
+        expandOnClick,
         loading_state,
+        selected,
         setProps,
         ...others
     } = props;
@@ -84,6 +113,14 @@ const Tree = (props: Props) => {
     useDidUpdate(() => {
         setProps({ checked: tree.checkedState });
     }, [tree.checkedState]);
+
+    useDidUpdate(() => {
+        setProps({ selected: tree.selectedState });
+    }, [tree.selectedState]);
+
+    useDidUpdate(() => {
+        tree.setSelectedState(selected);
+    }, [selected]);
 
     useDidUpdate(() => {
         expanded === "*"
@@ -110,7 +147,11 @@ const Tree = (props: Props) => {
             tree={tree}
             expandOnClick={false}
             renderNode={(payload) => (
-                <Leaf checkboxes={checkboxes} {...payload} />
+                <Leaf
+                    checkboxes={checkboxes}
+                    expandOnClick={expandOnClick}
+                    {...payload}
+                />
             )}
             {...others}
         />
@@ -119,6 +160,7 @@ const Tree = (props: Props) => {
 
 Tree.defaultProps = {
     expanded: [],
+    expandOnClick: true,
 };
 
 export default Tree;
--- a/src/ts/components/core/Tree.tsx
+++ b/src/ts/components/core/Tree.tsx
@@ -49,7 +49,6 @@ interface Props extends BoxProps, StylesApiProps, DashBaseProps {
 
 interface LeafProps {
     checkboxes: boolean;
-    expandOnClick: boolean;
     expandedIcon: React.ReactNode;
     collapsedIcon: React.ReactNode;
     iconSide?: "left" | "right" | "none";
@@ -63,7 +62,6 @@ const Leaf = (props: RenderTreeNodePayload & LeafProps) => {
         elementProps,
         tree,
         checkboxes,
-        expandOnClick,
         expandedIcon,
         collapsedIcon,
         iconSide,
@@ -86,17 +84,8 @@ const Leaf = (props: RenderTreeNodePayload & LeafProps) => {
                 : collapsedIcon}
         </span>
     );
-    const { onClick, ...otherElementProps } = elementProps;
     return (
-        <Group
-            gap="xs"
-            onClick={
-                expandOnClick
-                    ? () => tree.toggleExpanded(node.value)
-                    : undefined
-            }
-            {...otherElementProps}
-        >
+        <Group gap="xs" {...elementProps}>
             {iconSide === "left" && icon}
             {checkboxes && (
                 <Checkbox.Indicator
@@ -104,7 +93,7 @@ const Leaf = (props: RenderTreeNodePayload & LeafProps) => {
                     indeterminate={indeterminate}
                     onClick={(event) => {
                         event.stopPropagation();
-                        return !checked
+                        !checked
                             ? tree.checkNode(node.value)
                             : tree.uncheckNode(node.value);
                     }}
@@ -122,7 +111,6 @@ const Tree = (props: Props) => {
         checked,
         data,
         expanded,
-        expandOnClick,
         loading_state,
         selected,
         setProps,
@@ -135,6 +123,7 @@ const Tree = (props: Props) => {
     const tree = useTree({
         initialExpandedState: getTreeExpandedState(data, expanded),
         initialCheckedState: checked,
+        initialSelectedState: selected,
     });
 
     useDidUpdate(() => {
@@ -176,12 +165,10 @@ const Tree = (props: Props) => {
             }
             data={data}
             tree={tree}
-            expandOnClick={false}
             renderNode={(payload) => (
                 <Leaf
                     key={payload.node.value}
                     checkboxes={checkboxes}
-                    expandOnClick={expandOnClick}
                     expandedIcon={expandedIcon}
                     collapsedIcon={collapsedIcon}
                     iconSide={iconSide}
--- a/src/ts/components/core/Tree.tsx
+++ b/src/ts/components/core/Tree.tsx
@@ -179,6 +179,7 @@ const Tree = (props: Props) => {
             expandOnClick={false}
             renderNode={(payload) => (
                 <Leaf
+                    key={payload.node.value}
                     checkboxes={checkboxes}
                     expandOnClick={expandOnClick}
                     expandedIcon={expandedIcon}
--- a/src/ts/components/core/Tree.tsx
+++ b/src/ts/components/core/Tree.tsx
@@ -39,51 +39,79 @@ interface Props extends BoxProps, StylesApiProps, DashBaseProps {
     selected?: string[];
     /** Determines whether node should be selected on click, `false` by default */
     selectOnClick?: boolean;
+    /** Expanded state icon */
+    expandedIcon?: React.ReactNode;
+    /** Collapsed state icon */
+    collapsedIcon?: React.ReactNode;
+    /** Side to display expanded/collapsed state icon on, `'left'` by default */
+    iconSide?: "left" | "right" | "none";
 }
 
-const Leaf = ({
-    node,
-    expanded,
-    hasChildren,
-    elementProps,
-    tree,
-    checkboxes,
-    expandOnClick,
-}: RenderTreeNodePayload & { checkboxes: boolean; expandOnClick: boolean }) => {
+interface LeafProps {
+    checkboxes: boolean;
+    expandOnClick: boolean;
+    expandedIcon: React.ReactNode;
+    collapsedIcon: React.ReactNode;
+    iconSide?: "left" | "right" | "none";
+}
+
+const Leaf = (props: RenderTreeNodePayload & LeafProps) => {
+    const {
+        node,
+        expanded,
+        hasChildren,
+        elementProps,
+        tree,
+        checkboxes,
+        expandOnClick,
+        expandedIcon,
+        collapsedIcon,
+        iconSide,
+    } = props;
     const checked = tree.isNodeChecked(node.value);
     const indeterminate = tree.isNodeIndeterminate(node.value);
+    const icon = (
+        <span
+            style={{
+                visibility: hasChildren ? "visible" : "hidden",
+                transform:
+                    collapsedIcon !== undefined || expanded
+                        ? "rotate(0deg)"
+                        : "rotate(-90deg)",
+                transition: "0.2s",
+            }}
+        >
+            {collapsedIcon === undefined || expanded
+                ? expandedIcon
+                : collapsedIcon}
+        </span>
+    );
+    const { onClick, ...otherElementProps } = elementProps;
     return (
-        <Group gap="xs" {...elementProps}>
+        <Group
+            gap="xs"
+            onClick={
+                expandOnClick
+                    ? () => tree.toggleExpanded(node.value)
+                    : undefined
+            }
+            {...otherElementProps}
+        >
+            {iconSide === "left" && icon}
             {checkboxes && (
                 <Checkbox.Indicator
                     checked={checked}
                     indeterminate={indeterminate}
-                    onClick={() =>
-                        !checked
+                    onClick={(event) => {
+                        event.stopPropagation();
+                        return !checked
                             ? tree.checkNode(node.value)
-                            : tree.uncheckNode(node.value)
-                    }
+                            : tree.uncheckNode(node.value);
+                    }}
                 />
             )}
-            <Group
-                gap={5}
-                onClick={
-                    expandOnClick
-                        ? () => tree.toggleExpanded(node.value)
-                        : undefined
-                }
-            >
-                <span>{node.label}</span>
-                {hasChildren && (
-                    <AccordionChevron
-                        style={{
-                            transform: expanded
-                                ? "rotate(180deg)"
-                                : "rotate(0deg)",
-                        }}
-                    />
-                )}
-            </Group>
+            <span>{node.label}</span>
+            {iconSide === "right" && icon}
         </Group>
     );
 };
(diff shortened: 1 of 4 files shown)


AnnMarieW:
Hi @Godisemo 

Thanks so much for this PR!  :rocket:  

Nice example app too.  Here it is hosted on [PyCafe](https://py.cafe/snippet/dash/v1?#c=H4sIAM8aXmcEA91Y247bNhD9FcF92QC2vLL34jUgIG0KtH0o0oe8RYFAS5StNU2qFLWxd7H_3kNK9EqKLMsFiiKZILHJmTlz4QwdzssoEjEdLUfpLhNSOTHJN-GOcJVyGkYCm5xylTskd-JdFPBEip0RciqFP3hWqLHzsVDmMyKMrUi0HTu_QmjshAZQUh5TSWXAK63HXPAAf5psN8ypwopEKnyiMk8FvwpG3sKdudfB6J1WIFnm-Ab7yqxjogg2PgfcAb2UH5qCESMryoLREl9zGQWjcZ35RFhBTzGjTcpiOKX5FbKlmgVLDUtvKWtgWmobnp6V7_HF0kvDg5-jSMgYqXNVvgcotk7anLaEXztc0NS08ElSOgT8TW4Y7i-FUtaTfuS6ZBf2l9ZeXcby7F7tRBvecPRFuBNxwWjzbBqO9Uj1nFzNpqWGbdMCDThLJ41P-3R6XLHUPIoUDbl3Y2QYiFj3Gp3WpW1a29QRsqWG5QyXB1lT19wQXdFYOufUIKAub22BWGrLdETSiOB9dX12Gj3t9Rm1QUfYua2p4SBavtuKpfNOohnPYgzw2FKP55YaEdQLrscHSwPjadTxGdh2UbTp0ogGVWudhsZ0EXBfVO22sHRKpycDjcg3Qmz78z0g1CEg31U9moB-tIIsg7oIuS-s_6oiEyF3vY4NCHUAxndVjzqeH60cTUwXAfdFdUk1tmXrMpZn92qZGpaRRuSdUuegVR4JnqTrppamBna3mAb_Uj3VXEYOolB4oOHx6P5Z5v0vKZ5SvPeuSvla6Wuh36QosoplqSZiSYvq90VL0lIa-_AP_KNbbdIPR1__c4LP6BNlH5MEL1JA7ZGZE4J0nxH0Rex3eGmpWYEnoTQ1RY_F-m90hlxCpxWHdLut1TZFGxptV2JPc_-TLGiflEncUDe67L3r2NPV8QGQv6frDcNfdWUKwh4V4nH02ANb-isjfF2gSbAsK_kCyCqGSxHbgRBAckgnjO4nuSLy25fcWoqv7WxiBKI_LJra0B318ZLLZLoj8vBBMCHLXs1Svi0fZ2ZsEvD3dlJTdVA5wMG4pZ4kxIeg3tw_CtXCbsuYgRBEqu4D_w3xtMwRUIvAxZgmjuaGRzctChJdyr5blmCWgXFVTHHV6Iy7cbHL8iukwnJ1Guz317Gjq4or_6bKYAXZDXH0bWnlvgWQVBWSHy0YoKOnZlWmPU2cMORkR8PQ8X0EHmLalvIwBHqJpK9NWfCrmK6KtTlwmBiNMRP4u0glDhgDEMzr9OBMj7-6x3XvnY1SWb6cTrODG5GETtebKaoqTfTTOOcU__d4epxq9UmlPnlTn87mM2_hed7s1oh0WJhcu96t602yw3zCsTUh_OB-3TA4qg6ZHihqRayyg9oIjnV2EDGu_cnTtTu7cz2wyh-H0ZIXjI1HSYreHy0_fwFDRFu9_CC4gjEo_-RsAIcTfIgWseclq1lyk9zcR2Q2j-7niwV5WNzcX0dewFcMlU6l73vuA6aGPKI66tT3Z9ezGwwSZ9jaEIlrHX7LHRrvWUvP3blhsTTa-v7C9dz7MrtQdPUEslwhSbKeKc291mYMc6N2rIv5P51F5ZUiK0Z9_7b0JWEkR4RzrOYox5gTvUCw5WSWpavJjipifh6RB2iBpfIYNxrFT7MJSo9j-WPKH8lMK3saCffNtshyhKclPPc24JzmakLyA49Soc_jTqtlTCh20O7gOHBetix0jnwftaE3da1D2Ribe1oN_SUPKV9rnLl7E3AcYJEpIRiE7h7gJ9TydK_5njGEyiFRqmAKdWA2DrgE1xO6B0ePlqEIDwx6IRkCn5exAegrldtnivYzeboL-HOawTuUiBHPc-R29PoP4GSUVTwXAAA)

I'll take a closer look in the next couple days and get back to you with more comments and feedback.

In the meantime, since you obviously know a lot about both Mantine and Dash,  I hope you don't mind if I ask you a question about an open item in an  unrelated PR. (#458) that I'm working on right now:

- Is it possible to remove the `renderDashComponents` function from dash-extensions-js?   It's used to render components as props in a [few components](https://github.com/search?q=repo%3Asnehilvj%2Fdash-mantine-components%20dash-extensions-js&type=code) including the `Stepper`.   Is there a better way to do this?
 Here is the function as defined in  [`dash-extensions-js` ](https://github.com/emilhe/dash-extensions-js/blob/8b07ce3b61bbd3150cbf8dd6c5fb6d0204e74adf/index.js#L372).

Godisemo:
Thank you @AnnMarieW , unfortunately I don't have experience in the part of dash related to your other PR. I updated this PR with the rest of the upstream mantine component properties as well as updated the example app in the first comment. Let me know if there is anything else that is needed when you have had time to look at it.

AnnMarieW:
@Godisemo 

A Tree component like this has been a popular request and will make a great addition to Dash. :tada: 

When adding components to DMC, our goal is to be aligned as closely as possible with the features and prop names of the upstream Mantine component.    

It can be a little tricky to get this right. Let's take for example, the expanded feature:

Here, the `expanded` prop is defined as:
```
/** Determines expanded nodes as a list of values or `'*'` for all, `[]` by default */
expanded?: string[] | "*";
```

However the [Mantine Tree component](https://mantine.dev/core/tree/). has the following:

```
  /** `true` if the node is expanded, applicable only for nodes with `children` */
  expanded: boolean;

  /** A record of `node.value` and boolean values that represent nodes expanded state */
  expandedState: TreeExpandedState;
```
And  the following functions:
```
 /** Expands node with provided value */
  expand: (value: string) => void;

  /** Expands all nodes */
  expandAllNodes: () => void;

  /** Collapses all nodes */
  collapseAllNodes: () => void;
```

In the example app,  if you click on a node with no children it toggles on and off the `expanded` list, which is confusing.  It also might be better to have a separate prop for setting the expanded state. Currently, If you set `expanded='*'` either initially or in a callback, the `expanded` prop is updated with a list of values.

There are also similar features for the` checked` prop and they should be consistant.

It would be great to get some  feedback on the API, prop names etc from @alexcjohnson   I think it's best to take a little time with the design now so we don't have breaking changes later.

AnnMarieW:
@Godisemo - quick update.  Haven't forgotten about this PR.  Just getting caught up after the holidays and will get back to this soon. Thanks for your patience  :slightly_smiling_face:

alexcjohnson:
I think the prop names & behaviors are good. The `expanded="*"` behavior is a little funny, but I don't see a better alternative so I think we should just document it:
- If you set `expanded="*"`, when the component is rendered this will be replaced by the list of all expanded nodes. In the app running on py.cafe, the callback listening to `expanded` never seems to see the `"*"`, only the full list, which is probably what you'd want. But I'm not sure this would always be the case, for example if the tree was in an unrendered tab I bet you'd see the `"*"` until the user switches to the tree tab and the component is actually rendered. (Same is true if you put invalid entries in the `expanded` list, they're stripped out of the prop value when Mantine renders this.)
- A side effect of this, if you change the tree data and want to still have all nodes expanded, you'll need to again set `expanded="*"` along with the new tree data.

Also: even leaf nodes get listed in `expanded` (either by `expanded="*"` or by clicking on them) even though nothing changes visually. I would have called this a bug in this PR (ie don't call `tree.toggleExpanded(node.value)` unless `hasChildren`) if it were just about clicking them, but `tree.expandAllNodes()` seems to include them as well. To me that seems a bug in Mantine itself, but maybe I'm missing something.

src/ts/components/core/Tree.tsx
```diff
@@ -0,0 +1,166 @@
...
+                        : undefined
+                }
+            >
+                <span>{node.label}</span>
+                {hasChildren && (
+                    <AccordionChevron
+                        style={{
+                            transform: expanded
+                                ? "rotate(180deg)"
+                                : "rotate(0deg)",
```
alexcjohnson, line 82: I know this is how the Mantine docs do it, but I find this very confusing. The standard behavior, that you can see even right here in GitHub, is that a collapsed node has the chevron pointing to the right, and an expanded node has it pointing down.
Screenshot 2025-01-17 at 09 16 10
Which I think would be this:
```suggestion
                            transform: expanded
                                ? "rotate(0deg)"
                                : "rotate(270deg)",
```

AnnMarieW: Nice catch - I found those chevrons confusing but couldn't say why. lol

Part of the bigger issue with wrapping this component for Dash is how to customize things like this chevron, and formatting the Tree to include other components like icons.  Even this [forum topic](https://community.plotly.com/t/file-explorer-tree-generator-for-local-files/68732/) had a debate on whether the chevron should be to the right or left of the text....  

If you were creating an app using React and Mantine, then it's easy to customize all of these things.  But how can we design an API for Dash to allow this flexibility?

alexcjohnson: Right, that's always a challenge when going from a relatively low-level interface like Mantine to a higher-level interface like Dash components. Generally I'd say you _don't_ design for all the flexibility Mantine gives you; you make smart choices for the defaults and then add flexibility where a significant number of users will want it.

And "which side are the chevrons on" might be just such a case - personally I prefer them on the left, and given that that's where Mac, Windows, and GH all put them I think that would be a solid default, but since others like the right side we could certainly add a prop like `chevronSide = "left" | "right" | "none"`

One note though about putting them on the left: leaf nodes don't get a chevron, but we should probably add blank space as if they did, otherwise leaves won't look like they're really inside their parent folders.

alexcjohnson: Good point about the icon not necessarily being a chevron - so `chevronSide` would be limiting (also maybe not the most well-known word!). But do you ever see users wanting changeable icons on both sides? That feels weird to me, seems like the icon that changes when the node is opened or closed will either be on the left or right, and if you want an _unchanging_ icon on the other side you can presumably put that in the `label`?

If we agree on that, then we can have maybe `expandIconSide = "left" | "right" | "none"`?

Then to the question of changing the icons: The advantage of a CSS rotation is you can animate it (which means we should use `-90deg` rather than `270deg` as I had suggested above). Would animation still work if you specified the same icon with different CSS? Something like:
```
collapsedIcon=DashIconify(icon="tabler:chevron-down", style={"transform": "rotate(-90deg)", "transition": "transform 0.5s"})
expandedIcon=DashIconify(icon="tabler:chevron-down", style={"transform": "rotate(0deg)", "transition": "transform 0.5s"})
```
If that preserves the animation, it seems like a good compromise, covering the vast majority of use cases without bloating the API too much.

AnnMarieW: Changing the icons based on state, could be done like the checkbox with the checked or intermediate states:
https://www.dash-mantine-components.com/components/checkbox#change-icons

Godisemo:
> I think the prop names & behaviors are good. The `expanded="*"` behavior is a little funny, but I don't see a better alternative so I think we should just document it:
> 
>     * If you set `expanded="*"`, when the component is rendered this will be replaced by the list of all expanded nodes. In the app running on py.cafe, the callback listening to `expanded` never seems to see the `"*"`, only the full list, which is probably what you'd want. But I'm not sure this would always be the case, for example if the tree was in an unrendered tab I bet you'd see the `"*"` until the user switches to the tree tab and the component is actually rendered. (Same is true if you put invalid entries in the `expanded` list, they're stripped out of the prop value when Mantine renders this.)
> 
>     * A side effect of this, if you change the tree data and want to still have all nodes expanded, you'll need to again set `expanded="*"` along with the new tree data.
> 
> 
> Also: even leaf nodes get listed in `expanded` (either by `expanded="*"` or by clicking on them) even though nothing changes visually. I would have called this a bug in this PR (ie don't call `tree.toggleExpanded(node.value)` unless `hasChildren`) if it were just about clicking them, but `tree.expandAllNodes()` seems to include them as well. To me that seems a bug in Mantine itself, but maybe I'm missing something.

The functionality is indeed a bit quirky in upstream Mantine, but as you say I agree we shouldn't do anything about it here.

src/ts/components/core/Tree.tsx
Godisemo: > Good point about the icon not necessarily being a chevron - so `chevronSide` would be limiting (also maybe not the most well-known word!). But do you ever see users wanting changeable icons on both sides? That feels weird to me, seems like the icon that changes when the node is opened or closed will either be on the left or right, and if you want an _unchanging_ icon on the other side you can presumably put that in the `label`?
> 
> If we agree on that, then we can have maybe `expandIconSide = "left" | "right" | "none"`?
> 
> Then to the question of changing the icons: The advantage of a CSS rotation is you can animate it (which means we should use `-90deg` rather than `270deg` as I had suggested above). Would animation still work if you specified the same icon with different CSS? Something like:
> 
> ```
> collapsedIcon=DashIconify(icon="tabler:chevron-down", style={"transform": "rotate(-90deg)", "transition": "transform 0.5s"})
> expandedIcon=DashIconify(icon="tabler:chevron-down", style={"transform": "rotate(0deg)", "transition": "transform 0.5s"})
> ```
> 
> If that preserves the animation, it seems like a good compromise, covering the vast majority of use cases without bloating the API too much.

Animation does indeed still work when specifying the same icon. I have updated the PR to include this functionality.

Godisemo:
I have updated the PR to include three new properties `expandedIcon`, `collapsedIcon` and `iconSide`. By default icon side is left and the expanded icon is the accordeon chevron included in mantine. When no collapsed icon is specified it is assumed that the expanded icon should be rotated into the collapsed state. When a collapsed icon is specified in addition to an expanded icon no transform will be applied.

AnnMarieW:
The changes look great!  :tada: 

Here's a draft  of the docs:  https://github.com/snehilvj/dmc-docs/pull/147

For the `expanded` states weirdness, I included the following in the docs - Is it clear enough?

> ### Expanded State in callbacks.
>
>When using the `expanded` property as a callback input to track the user's selected expanded state, note that the expanded
list may include or exclude leaf nodes (nodes without children) depending on user interaction.
>
>This happens because users can toggle the state of leaf nodes, even though they don’t affect how the tree data is 
displayed. To handle this, ensure your callback logic accounts for the possibility that leaf nodes may or may not
be present in the `expanded` prop.

I only found one issue:   Not sure what's causing this error in the console: 
> Warning: Each child in a list should have a unique "key" prop

@Godisemo are you familiar with Dash testing?  If so, you can check out the tests for other components in the `tests` folder.  We just need tests for Dash specific props and Dash callback.  Other props that are just passed through to Mantine don't need a test.  Let me know if you would like me to help with this.  It can be a little tricky to set up if you haven't done it before.

Godisemo:
> I only found one issue: Not sure what's causing this error in the console:
> 
> > Warning: Each child in a list should have a unique "key" prop

I also noticed this, but I concluded it comes from the code highlighting component.

> @Godisemo are you familiar with Dash testing? If so, you can check out the tests for other components in the tests folder. We just need tests for Dash specific props and Dash callback. Other props that are just passed through to Mantine don't need a test. Let me know if you would like me to help with this. It can be a little tricky to set up if you haven't done it before.

I will try to add some unit tests.

AnnMarieW:
The console error is not coming from the `CodeHightlight`  component.  You can see it even in this basic example:

```python

app = Dash()


data2 = [
    {
        "value": "src",
        "label": "src",
        "children": [
            {"value": "src/components", "label": "components"},
            {"value": "src/hooks", "label": "hooks"},
        ],
    },
    {"value": "package.json", "label": "package.json"},
]


app.layout = dmc.MantineProvider(dmc.Tree(data=data))

if __name__ == "__main__":
    app.run(debug=True)





```

AnnMarieW:
@Godisemo 
I'm not looking to add feature creep to this PR, but can you tell me if it's possible to make the label a React node instead of just a string?  It would make it much easier to format the text, add icons, or even make it a link.  This would be a very useful component for navigation if it supported links.

Godisemo:
> @Godisemo I'm not looking to add feature creep to this PR, but can you tell me if it's possible to make the label a React node instead of just a string? It would make it much easier to format the text, add icons, or even make it a link. This would be a very useful component for navigation if it supported links.

@AnnMarieW It might be possible to do, but I think it would require making some sort of custom leaf renderer dmc component.

AnnMarieW:
> It might be possible to do, but I think it would require making some sort of custom leaf renderer dmc component.

OK, that makes sense.  That could be a path forward for a future PR if this is a popular request.

Godisemo:
@AnnMarieW I have added a few tests. Let me know if there is something else that should be updated before this PR can be accepted.

AnnMarieW:
@Godisemo 
This looks great!  Nice tests !   Now just need a changelog entry.

@alexcjohnson - do you have any more comments before merging?

tests/test_tree.py
```diff
@@ -0,0 +1,176 @@
...
+                html.Button(id="button-clear-data", children="clear data"),
+                html.Button(id="button-clear-selection", children="clear selection"),
+                html.Button(id="button-clear-checked", children="clear checked"),
+            ]
+        )
+    )
+
+    @app.callback(Output("output-checked", "children"), Input("tree", "checked"))
+    def update_output_checked(checked):
+        return str(checked)
```
alexcjohnson, line 48: When I was playing with this feature I found the ordering of lists returned by the tree to be confusing, as it depends in odd ways on the order of operations. So although the tests as you've written them are robust for the current version of Mantine, I wouldn't be surprised if in the future they change this. Also the tests themselves would be easier to read if these fields were sorted - which as you've constructed it is also the order on screen.

So I'd suggest changing this line to:

```suggestion
        return str(sorted(checked) if checked else checked)
```
(the `if checked else checked` just lets `None` through without an error, not sure if that's necessary) and the same for `expanded` and `selected` (not `data` of course 😄) and adjusting the test assertions to match.

And perhaps worth mentioning this inconsistent order in the docs?

alexcjohnson (approved):
💃 This looks great to me. Lovely tests 😻 Just one suggestion but that's not blocking.

Godisemo:
> 💃 This looks great to me. Lovely tests 😻 Just one suggestion but that's not blocking.

Good point regather tests, I’d be happy to update them before merging this.

Godisemo:
Updated the tests now. @AnnMarieW Thank you for adding the docs for this. Will you also mention what Alex pointed out about the `expanded`, `selected` and `checked` properties being sorted depending on order of operations?

AnnMarieW:
> Will you also mention what Alex pointed out about the expanded, selected and checked properties being sorted depending on order of operations?

Yes, I'll update the docs to include this.  

Thanks so much for adding this component and `InlineCodeHighlight`.  I'm looking forward to the next release - which should be in the next day or two. :partying_face:

PR #491, opened 2025-01-24
Add input wrapper component

Godisemo:
This enables you to write custom inputs with consistent styling:

bild

Example app:
```python

    dmc.Flex(
        align="center",
        justify="center",
        style={"background": "lightgray", "height": "100vh", "width": "100vw"},
        children=dmc.Card(
            w=400,
            children=[
                dmc.TextInput(
                    label="Text input",
                    description="This is a text input",
                    error="Don't write that!",
                    withAsterisk=True,
                ),
                dmc.InputWrapper(
                    label="Tree input",
                    description="This is a tree input",
                    error="Don't cut down trees!",
                    inputWrapperOrder=["label", "description", "error", "input"],
                    withAsterisk=True,
                    children=[
                        dmc.Tree(
                            checkboxes=True,
                            data=[
                                {
                                    "label": "root",
                                    "value": "value",
                                    "children": [
                                        {"label": "child 1", "value": "child_1"},
                                        {"label": "child 2", "value": "child_2"},
                                    ],
                                }
                            ],
                        )
                    ],
                ),
                dmc.NumberInput(
                    label="Number input",
                    description="This is a number input",
                    error="Not that number!",
                ),
            ],
        ),
    )
)


if __name__ == "__main__":
    app.run(debug=True)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,154 +1,160 @@
 # Change Log
 
+# Unreleased
+
+### Added
+
+-   Added `InputWrapper` component #491 by @Godisemo
+
 # 0.15.2
 
 ### Added
-- Added `disabled` prop for individual options on a `SegmentedControl`  #451  by @petefine 
-- Added `type` prop for `SimpleGrid` to enable container queries #472 by @namakshenas
-- Added `type` prop and `breakpoints` prop for `Grid` to enable container queries #473 by @namakshenas
-- Added props #478 by @AnnMarieW:
-    - `inline` prop to `Center`, 
-    - `allowedInitialsColors` to `Avatar`
-    - `defaultExpanded` to `CodeHightlightTabs`
-    - `autoComplete` to `PasswordInput`
-  
-- Added props available in Mantine 7.16.0 #489 by @AnnMarieW
-    - `withOverlay` and `overlayProps` to `Popover
-    - `withPages` to `Pagination
-    - `type` to `Carousel` enabling container queries
-    - `restrictToMarks` to `RangeSlider`
-  
 
-- Added `Tree` component #460 by @Godisemo
-- Added `InlineCodeHighlight` component #483 by @Godisemo
+-   Added `disabled` prop for individual options on a `SegmentedControl` #451 by @petefine
+-   Added `type` prop for `SimpleGrid` to enable container queries #472 by @namakshenas
+-   Added `type` prop and `breakpoints` prop for `Grid` to enable container queries #473 by @namakshenas
+-   Added props #478 by @AnnMarieW:
+    -   `inline` prop to `Center`,
+    -   `allowedInitialsColors` to `Avatar`
+    -   `defaultExpanded` to `CodeHightlightTabs`
+    -   `autoComplete` to `PasswordInput`
+-   Added props available in Mantine 7.16.0 #489 by @AnnMarieW
+
+    -   `withOverlay` and `overlayProps` to `Popover
+    -   `withPages` to `Pagination
+    -   `type` to `Carousel` enabling container queries
+    -   `restrictToMarks` to `RangeSlider`
+
+-   Added `Tree` component #460 by @Godisemo
+-   Added `InlineCodeHighlight` component #483 by @Godisemo
 
 ### Fixed
-- Styled the figure buttons in the "mantine_light" and "mantine_dark" Plotly figure templates.  #456 by @AnnMarieW
-- In `Carousel`  Allow only "center" | "start" | "end" in the `align` prop #477 by @AnnMarieW
+
+-   Styled the figure buttons in the "mantine_light" and "mantine_dark" Plotly figure templates. #456 by @AnnMarieW
+-   In `Carousel` Allow only "center" | "start" | "end" in the `align` prop #477 by @AnnMarieW
 
 ### Changed
-- Upgraded to latest Mantine (7.16.1)
 
+-   Upgraded to latest Mantine (7.16.1)
 
 # 0.15.1
 
 ### Added
 
-- Added `disabled` prop to `Radio`  #425  by @namakshenas
-- Added `lineSize` prop to `Burger`  #432  by @ChinoUkaegbu
-- Added `SemiCircleProgress` component #434 by @AnnMarieW
+-   Added `disabled` prop to `Radio` #425 by @namakshenas
+-   Added `lineSize` prop to `Burger` #432 by @ChinoUkaegbu
+-   Added `SemiCircleProgress` component #434 by @AnnMarieW
 
-- Added the `add_figure_templates()` function which creates Mantine-styled Plotly figure templates for
-both light and dark modes using the default Mantine theme.  It registers the templates with plotly.io.templates as
-"mantine_light" and "mantine_dark", and optionally sets one of the themes as a default. #431 by @AnnMarie
+-   Added the `add_figure_templates()` function which creates Mantine-styled Plotly figure templates for
+    both light and dark modes using the default Mantine theme. It registers the templates with plotly.io.templates as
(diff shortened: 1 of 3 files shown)


AnnMarieW:
Hi @Godisemo 
This looks great!  Thanks for wrapping another component.  

I can see how this would be useful.  Depending on how it's used, accessibility could be an issue.  In Dash you can't set the wrapper and the input (child component) to the same id, so it's not possible to link them.  https://mantine.dev/core/input/#accessibility

Godisemo:
That is a good point, but actually when I tried it now I don't get any error messages and everything works as expected despite setting id to the same value for the wrapper and the child. See this updated example:

```python
from dash import Dash, _dash_renderer, Input, Output, html, dcc

    dmc.Flex(
        align="center",
        justify="center",
        style={"height": "100vh", "width": "100vw"},
        children=dmc.Card(
            w=400,
            children=[
                dmc.TextInput(
                    label="Text input",
                    description="This is a text input",
                    error="Don't write that!",
                    withAsterisk=True,
                ),
                dmc.InputWrapper(
                    id="tree",
                    label="Tree input",
                    description="This is a tree input",
                    error="Don't cut down trees!",
                    inputWrapperOrder=["label", "description", "error", "input"],
                    withAsterisk=True,
                    children=[
                        dmc.Tree(
                            id="tree",
                            checkboxes=True,
                            data=[
                                {
                                    "label": "root",
                                    "value": "value",
                                    "children": [
                                        {"label": "child 1", "value": "child_1"},
                                        {"label": "child 2", "value": "child_2"},
                                    ],
                                }
                            ],
                        )
                    ],
                ),
                dmc.InputWrapper(
                    id="unstyled",
                    label="Unstyled input",
                    description="This is an unstyled input",
                    error="Stop it!",
                    children=[dcc.Input(id="unstyled")],
                ),
                dmc.NumberInput(
                    id="number",
                    label="Number input",
                    description="This is a number input",
                    error="Not that number!",
                ),
            ],
        ),
    ),
    forceColorScheme="dark",
)


@app.callback(Output("number", "value"), Input("tree", "checked"))
def update_checked(checked: list[str]):
    return len(checked)


if __name__ == "__main__":
    app.run(debug=True)

```

Godisemo:
> That is a good point, but actually when I tried it now I don't get any error messages and everything works as expected despite setting id to the same value for the wrapper and the child.

I guess this could be because setting the id on the wrapper doesn't actually set the id on the underlying mantine components, but instead sets the `for` attribute on the label. It is slightly confusing so it would need to be documented on the DMC side for sure.

AnnMarieW:
Huh, when I run that example, I get this error:
> DuplicateIdError
dash.exceptions.DuplicateIdError: Duplicate component id found in the initial layout: `tree`

AnnMarieW:
I wonder if we can just add an `htmlFor` so it can be  set manually if needed.   I'm thinking about the case where someone wraps something like a `dcc.Dropdown` with the `InputWrapper`

 But I'm no accessibility expert,so I don't know how that would work.  Just thinking about if someone wrapped a `dmc.Checkbox` with the `InputWrapper`  Would that mess up the label that's already tied to the `Checkbox`

Godisemo:
> Huh, when I run that example, I get this error:

Now I see, I didn't get this at first, but it does indeed show up the first time the app is loaded fresh. On refresh the error disappears.

Another problem is that even if using duplicate ids would have worked for associating the label with the input, it stops you from being able to set e.g. the error message in a callback since the input component (which doesn't have the error property) would be referenced rather than the wrapper component.

I like your idea with adding a new property of htmlFor to the wrapper component. This can then be passed to the underlying mantine id property. The dash id property will not be passed on to the mantine component but will instead just be used to reference the dash component in e.g. callbacks.

I will make a new commit for this with another updated example app.

Godisemo:
> I'm thinking about the case where someone wraps something like a `dcc.Dropdown` with the `InputWrapper`

I cannot get the `htmlFor` attribute of `html.Label` to play nice with `dcc.Dropdown` in pure dash. I would have expected that clicking the label would open the dropdown. Perhaps there is still a value in addin the `htmlFor` attribute for accessibility.

Godisemo:
updated example:
```python
from dash import Dash, _dash_renderer, Input, Output, dcc

    dmc.Flex(
        align="center",
        justify="center",
        style={"height": "100vh", "width": "100vw"},
        children=dmc.Card(
            w=400,
            children=[
                dmc.TextInput(
                    id="text",
                    label="Text input",
                    description="This is a text input",
                    withAsterisk=True,
                ),
                dmc.InputWrapper(
                    id="tree-wrapper",
                    label="Tree input",
                    description="This is a tree input",
                    inputWrapperOrder=["label", "description", "error", "input"],
                    withAsterisk=True,
                    children=[
                        dmc.Tree(
                            id="tree",
                            checkboxes=True,
                            data=[
                                {
                                    "label": "root",
                                    "value": "value",
                                    "children": [
                                        {"label": "child 1", "value": "child_1"},
                                        {"label": "child 2", "value": "child_2"},
                                    ],
                                }
                            ],
                        )
                    ],
                ),
                dmc.InputWrapper(
                    id="unstyled-wrapper",
                    htmlFor="unstyled",
                    label="Unstyled input",
                    description="This is an unstyled input",
                    children=[dcc.Input(id="unstyled")],
                ),
                dmc.NumberInput(
                    id="number",
                    label="Number input",
                    description="This is a number input",
                    error="Not that number!",
                ),
            ],
        ),
    ),
    forceColorScheme="dark",
)


@app.callback(
    Output("text", "error"),
    Output("tree-wrapper", "error"),
    Output("unstyled-wrapper", "error"),
    Output("number", "error"),
    Input("text", "value"),
    Input("tree", "checked"),
    Input("unstyled", "value"),
    Input("number", "value"),
)
def validate(text: str, tree: list[str], unstyled: str, number: int):
    text_error = "MUST BE ALL CAPS" if text and text != text.upper() else None
    tree_error = "Select at least one" if not tree else None
    unstyled_error = (
        "Must be longer than 10 characters" if unstyled and len(unstyled) < 10 else None
    )
    number_error = "Must be larger than 10" if number and number <= 10 else None
    return text_error, tree_error, unstyled_error, number_error


if __name__ == "__main__":
    app.run(debug=True)

```

AnnMarieW:
Thanks for the update @Godisemo 

@alexcjohnson do you know of someone who can check this for accessibility?

AnnMarieW:
@Godisemo I think that the htmlFor not working with dcc.Dropdown it because that dcc component is not accessible.  It works well with the `dcc.Input`.  You can see that it focuses the dcc.Input when the label is clicked.

I think this is good to go.  I'll just see if Alex has any more feedback before merging.

Thanks again!

alexcjohnson (approved):
💃 Looks good. I'll note that `htmlFor` here won't work for pattern-matching IDs - or rather, you'd have to explicitly construct the stringified form of the ID, which I don't believe is a process we expose or document. I feel like this has come up with `htmlFor` attributes before, but I can't find it now. Anyway that's a bit of an edge case so I don't mind leaving it until someone needs it. For `id` props this happens in the Dash renderer so components themselves never need to worry about it, so the proper solution may be to create a mechanism in the renderer to designate props other than `id` as needing ID stringification.

I'm not that familiar with the mechanisms behind `for`, so I don't know if there would be a way to make it work with JS-driven components like `dcc.Dropdown`, but feel free to open an issue - or better yet a PR if you figure it out - in the Dash repo.

AnnMarieW:
Thanks @alexcjohnson All great points!
We can document the limitations for now.  I'm familiar with how to stringify the dict ids for pattern matching, but I haven't tried it with `htmlFor`.  I'll open an issue and try to add that in another PR.

PR #471, opened 2025-01-07
Allow 'debounce=True' in MonthPickerInput

oelhammouchi:
This PR closes #436, it basically just adapts #405. Here's an example app, the callback will never fire with just one end of the range set.

```python
from datetime import date
from typing import Tuple

from dash import Dash, Input, Output, _dash_renderer, callback, html

import dash_mantine_components as dmc

_dash_renderer._set_react_version("18.2.0")

app = Dash(__name__, external_stylesheets=dmc.styles.ALL)

app.layout = dmc.MantineProvider(
    [
        html.H1(children="My beautiful app", style={"textAlign": "center"}),
        dmc.Container(
            dmc.Stack(
                [
                    dmc.MonthPickerInput(
                        id="month-range", type="range", debounce=True, clearable=True
                    ),
                    dmc.Text(id="show-range"),
                ],
            )
        ),
    ]
)


@callback(
    Output("show-range", "children"),
    Input("month-range", "value"),
    prevent_initial_call=True,
)
def month_range_changed(month_range: Tuple[date, date]):
    if month_range is not None:
        return f"Between {month_range[0]} and {month_range[1]}"


if __name__ == "__main__":
    app.run(debug=True)

```

PR #496, opened 2025-01-28
Fix debounce by updating popover returnFocus

### LINKED ISSUE #495: Debounce stopped working in DatePickerInput in 0.15.2
When `debounce=True`, the value is updated when the input loses focus.  

In a recent [Mantine update](https://github.com/mantinedev/mantine/commit/7dbda1e6f2c95aaf1e52251b1874c15de2e59872#diff-8e830fef686217ced24afe002b543505b26c944ca69a51d5b840e156664304f1R164), they changed the default popover prop  `returnFocus` to false,  which meant that the Input wasn't focused after the dates were selected.  This broke the debounce because it wasn't possible to update the value when the input loses focus. 

```python
from dash import Dash, _dash_renderer, Input, Output, html, callback

from datetime import datetime, date, timedelta


component = html.Div(
    [
        dmc.DatePickerInput(
            id="date-picker-input-multiple",
            label="Pick dates",
            description="Pick one or more dates",
            minDate=date(2020, 8, 5),
            value=[datetime.now().date(), datetime.now().date() + timedelta(days=5)],
            w=400,
            type="multiple",
            placeholder="Pick dates",
            maw=300,
            clearable=True,
            debounce=True,
            popoverProps={"returnFocus": True}   # this is now required for the debounce to work.
        ),
        dmc.Space(h=10),
        dmc.Text(id="selected-date-input-multiple"),
    ]
)


@callback(
    Output("selected-date-input-multiple", "children"),
    Input("date-picker-input-multiple", "value"),
)
def update_output(dates):
    return str(dates)

    component
)

if __name__ == "__main__":
    app.run(debug=True)




```

AnnMarieW:
closes #495 

When `debounce=True`, the value is updated when the input loses focus.  

In a recent [Mantine update](https://github.com/mantinedev/mantine/commit/7dbda1e6f2c95aaf1e52251b1874c15de2e59872#diff-8e830fef686217ced24afe002b543505b26c944ca69a51d5b840e156664304f1R164), they changed the default popover prop  `returnFocus` to false,  which meant that the Input wasn't focused after the dates were selected.  This broke the debounce because it wasn't possible to update the value when the input loses focus. 

```python
from dash import Dash, _dash_renderer, Input, Output, html, callback

from datetime import datetime, date, timedelta


component = html.Div(
    [
        dmc.DatePickerInput(
            id="date-picker-input-multiple",
            label="Pick dates",
            description="Pick one or more dates",
            minDate=date(2020, 8, 5),
            value=[datetime.now().date(), datetime.now().date() + timedelta(days=5)],
            w=400,
            type="multiple",
            placeholder="Pick dates",
            maw=300,
            clearable=True,
            debounce=True,
        #    popoverProps={"returnFocus": True}   # this was required in order for the debounce to work in 0.15.2
        ),
        dmc.Space(h=10),
        dmc.Text(id="selected-date-input-multiple"),
    ]
)


@callback(
    Output("selected-date-input-multiple", "children"),
    Input("date-picker-input-multiple", "value"),
)
def update_output(dates):
    return str(dates)

    component
)

if __name__ == "__main__":
    app.run(debug=True)




```

### DIFF
--- a/src/ts/components/dates/DatePickerInput.tsx
+++ b/src/ts/components/dates/DatePickerInput.tsx
@@ -101,6 +101,7 @@ const DatePickerInput = (props: Props) => {
                 minDate={stringToDate(minDate)}
                 maxDate={stringToDate(maxDate)}
                 excludeDate={isExcluded}
+                popoverProps={{returnFocus: true}}
                 {...others}
             />
         </div>
--- a/src/ts/components/dates/DatePickerInput.tsx
+++ b/src/ts/components/dates/DatePickerInput.tsx
@@ -41,6 +41,7 @@ const DatePickerInput = (props: Props) => {
         minDate,
         maxDate,
         disabledDates,
+        popoverProps,
         persistence,
         persisted_props,
         persistence_type,
@@ -101,7 +102,7 @@ const DatePickerInput = (props: Props) => {
                 minDate={stringToDate(minDate)}
                 maxDate={stringToDate(maxDate)}
                 excludeDate={isExcluded}
-                popoverProps={{returnFocus: true}}
+                popoverProps={{returnFocus: true, ...popoverProps}}
                 {...others}
             />
         </div>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+
+# Unreleased
+
+### Fixed
+
+- fixed debounce on DatePickerInput that stopped working in 0.15.2  #496 by @AnnMarie
+
 # 0.15.2
 
 ### Added


src/ts/components/dates/DatePickerInput.tsx
```diff
@@ -101,6 +101,7 @@ const DatePickerInput = (props: Props) => {
                 minDate={stringToDate(minDate)}
                 maxDate={stringToDate(maxDate)}
                 excludeDate={isExcluded}
+                popoverProps={{returnFocus: true}}
                 {...others}
```
alexcjohnson, line 105: Would we want to merge this with any `popoverProps` you provide explicitly? ie pull `popoverProps` out of `props`, then here:
```
popoverProps={{returnFocus: true, ...popoverProps}}
```

AnnMarieW: oh nice catch - yes, that's how it should work.

AnnMarieW: That needs to be applied to #471 too

alexcjohnson (approved):
💃

PR #504, opened 2025-02-05
NavLink Update for Active

### LINKED ISSUE #365: [Feature Request] Add `active` prop to `NavLink`
It would be great to have an `active` prop  on `NavLink` similar to Dash Bootstrap Components 
- [dbc.NavLink docs](https://dash-bootstrap-components.opensource.faculty.ai/docs/components/nav/)
- [dbc.Navlink GitHub]( https://github.com/facultyai/dash-bootstrap-components/blob/main/src/components/nav/NavLink.js)

> `active` (boolean | a value equal to: 'partial', 'exact'; default False): Apply 'active' style to this component. Set to "exact" to automatically toggle active status when the current pathname matches href, or to "partial" to automatically toggle on a partial match. Assumes that href is a relative url such as /link rather than an absolute such as https://example.com/link For example - dbc.NavLink(..., href="/my-page", active="exact") will be active on "/my-page" but not "/my-page/other" or "/random" - dbc.NavLink(..., href="/my-page", active="partial") will be active on "/my-page" and "/my-page/other" but not "/random".

This makes it possible to show which link in a menu or side panel is active regardless of how you navigate to the page.

BSd3v:
adding support for `NavLink` to set active based upon page navigation automatically:

- options for `active` are now `boolean | 'partial' | 'exact'`
- partial will care if the pathname starts with the `href`
- exact will display if the pathname is an exact match
- true / false will disable the matching and set true and false respectively

closes #365

### DIFF
--- a/src/ts/components/core/NavLink.tsx
+++ b/src/ts/components/core/NavLink.tsx
@@ -79,7 +79,7 @@ const NavLink = ({
         setLinkActive(
           active === true ||
             (active === 'exact' && pathname === href) ||
-            (active === 'partial' && (pathname.startsWith(href + '/') || pathname == href))
+            (active === 'partial' && (pathname.startsWith(href + '/') || pathname === href))
         );
     };
 
--- a/src/ts/components/core/NavLink.tsx
+++ b/src/ts/components/core/NavLink.tsx
@@ -27,7 +27,7 @@ interface Props
      mimics NavLink behaviour from dash-bootstrap-components
      exact will match the pathname exactly, whereas partial will only be concerned about the startsWith
      */
-    active?: boolean | 'partial' | 'exact';
+    active?: boolean | 'partial' | 'exact' | 'hyperlink';
     /** Key of `theme.colors` of any valid CSS color to control active styles, `theme.primaryColor` by default */
     color?: MantineColor;
     /** href */
@@ -77,16 +77,27 @@ const NavLink = (props: Props) => {
         setLinkActive(
           active === true ||
             (active === 'exact' && pathname === href) ||
-            (active === 'partial' && pathname.startsWith(href))
+            (active === 'partial' && pathname.startsWith(href)) ||
+            (active === 'hyperlink' && pathname.endsWith(href))
         );
     };
 
+    const parsePath = (location) => {
+        if (active === 'hyperlink') {
+          pathnameToActive(location.href)
+        }
+        else {
+          pathnameToActive(location.pathname)
+        }
+    }
+
     useEffect(() => {
-        pathnameToActive(window.location.pathname);
+        parsePath(window.location);
 
         if (typeof active === 'string') {
           History.onChange(() => {
-            pathnameToActive(window.location.pathname);
+              parsePath(window.location);
+            ;
           });
         }
     }, [active]);
--- a/src/ts/components/core/NavLink.tsx
+++ b/src/ts/components/core/NavLink.tsx
@@ -26,12 +26,11 @@ interface Props
     /** Section displayed on the right side of the label */
     rightSection?: React.ReactNode;
     /**
-     * Determines whether the link should have active styles, `false` by default. Can also set active styles based on the URL match type:
-     * - `exact`: The link is active only if the `pathname` exactly matches `href`.
-     * - `starts-with`: The link is active if the `pathname` begins with `href`, allowing for subpages.
-     * - `ends-with`: The link is active if the full URL (including query parameters and hash) ends with `href`
+     * Controls whether the link is styled as active (default: `false`).
+     * - `exact`: Active if `pathname` matches `href` exactly.
+     * - `partial`: Active if `pathname` starts with `href` (for subpages).
      */
-    active?: boolean | 'starts-with' | 'exact' | 'ends-with';
+    active?: boolean | 'exact' | 'partial';
     /** Key of `theme.colors` of any valid CSS color to control active styles, `theme.primaryColor` by default */
     color?: MantineColor;
     /** href */
@@ -71,7 +70,7 @@ const NavLink = ({
     persistence_type,
     setProps,
     loading_state,
-    active,
+    active = false,
     ...others
 }: Props) => {
     const [linkActive, setLinkActive] = useState(false);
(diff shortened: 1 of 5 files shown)


AnnMarieW:
Thanks @BSd3v 

This works great!  Will target merging  after dmc V1 is released (and after dash 3 is released).

To do:
- Updates for V1
- Changelog
- docs
- update `active` docstring

I think the dosctring could be a little more clear.  Perhaps something like:

>  Apply 'active' style to the link. Set to "exact" to automatically toggle the active status when the current pathname matches href, or to  "partial" to automatically toggle when the pathname starts with the href. 

I like the example included in the dbc docstring, but it would be better to include it in the docs instead:

 >For example
    - dbc.NavLink(..., href="/my-page", active="exact") will be active on
      "/my-page" but not "/my-page/other" or "/random"
    - dbc.NavLink(..., href="/my-page", active="partial") will be active on
      "/my-page" and "/my-page/other" but not "/random"
   


Here's a good sample app for the release announcement and the docs.  This demos nested links using the `active="partial"` and `active="exact"`  to keep the parent link showing active when any of the child links are active.


image

image


```python
import dash
from dash import Dash, _dash_renderer, html

app = Dash(external_stylesheets=dmc.styles.ALL, use_pages=True, pages_folder="")

dash.register_page("home", path="/", layout=html.Div("I'm home"))
dash.register_page("page1", path="/page-1", layout=html.Div("Info about page 1 subjects"))
dash.register_page("page1s1", path="/page-1/sub-1", layout=html.Div("page 1 subject 1"))
dash.register_page("page1s2", path="/page-1/sub-2", layout=html.Div("page 1 subject 2"))

component = dmc.Box([
    dmc.NavLink(label="home", href="/", active="exact"),
    dmc.NavLink(
            label="Page 1",
            childrenOffset=28,
            href="/page-1",
            active="partial",
            children=[
                dmc.NavLink(label="Subject 1", href="/page-1/sub-1", active="exact"),
                dmc.NavLink(label="Subject 2", href="/page-1/sub-2", active="exact"),
            ],
    ),
    dmc.Divider(mb="lg"),
    dash.page_container
])


if __name__ == "__main__":
    app.run(debug=True)


```

src/ts/components/core/NavLink.tsx
```diff
@@ -64,9 +69,39 @@ const NavLink = (props: Props) => {
...
+    };
+
+    const parsePath = (location) => {
+        if (active === 'hyperlink') {
+          pathnameToActive(location.href)
+        }
+        else {
+          pathnameToActive(location.pathname)
+        }
+    }
```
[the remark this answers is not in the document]
BSd3v: Sure, I think that would be fine.

Was more taking the lead from dbc on it.

AnnMarieW:
Hey @alexcjohnson 
This PR still needs to be updated for V1, but what do you think about the concept and prop names - especially the option for matching the href that includes hash tags and query strings.

I tried it out in the dmc docs and it works great.  Eliminates the need to do the clientside callback to highlight the active links.

alexcjohnson:
I like it! It's going a bit beyond just wrapping Mantine, but it seems like it's for a good cause.

Be careful about the logic though - just `startsWith` and `endsWith` would give strange results, like if you have two top-level path parts where one word is the beginning of the other - a word and its plural for example but there are plenty of otherwise unrelated word pairs that would do this. I think we'd want `starts-with-path` to only match if either the path is exact or the next character after the match is a `/`. `ends-with-url` I guess if people always start their `href` at either `?` or `#` it doesn't have that same danger, so we could opt to just document that as a recommendation but if someone wants to ignore that we leave it up to them. I might also suggest a flavor for exact including query params and hash. Not full URL, I think we still want to start with the path, so the naming is tricky, but maybe `exact-path-plus`?

BSd3v:
@alexcjohnson,

The thinking behind the `endsWith` is that this could be used to highlight links within the document as they are activated or clicked on. Similar to how this is done here:

https://www.ag-grid.com/javascript-data-grid/grid-events/#reference-miscellaneous

If you have to match the exact path, the links become much more of a pain.

We could match an `exact-search`? to be an exact `href`. Especially if someone was to use layouts in this fashion. 😄

alexcjohnson:
I wasn’t suggesting removing `ends-with-url`, just being clear on how to use it to avoid unexpected behavior, ie recommending always starting with `#` or `?`, or I guess `/`.

AnnMarieW:
Reverted to include just `active="exact"` and `active="partial"`

This now matches the functionality and prop names in the dash-bootstrap-components library.  It's simpler and will be easier to maintain and document.  

Draft of the docs are here:  https://github.com/snehilvj/dmc-docs/pull/165

src/ts/components/core/NavLink.tsx
```diff
@@ -64,8 +70,28 @@ const NavLink = ({
...
+    active = false,
     ...others
 }: Props) => {
+    const [linkActive, setLinkActive] = useState(false);
+
+    const pathnameToActive = pathname => {
+        setLinkActive(
+          active === true ||
+            (active === 'exact' && pathname === href) ||
+            (active === 'partial' && pathname.startsWith(href))
```
alexcjohnson, line 82: I still think `startsWith` is going to lead to unintended consequences, and `partial` should instead use
```js
pathname === href || pathname.startsWith(href + '/')
```
To account for paths like:
```
/al
/al/subpage
/alex
/alex/subpage
```
If you use just `startsWith`, then `href='/al'` would match all of these, but you really only want to match the first two.

BSd3v: What if we split at '/' and made sure that all of the link href is included in the same of the href. 

I feel like that may be stronger.

AnnMarieW: if you do this
```
pathname === href || pathname.startsWith(href + '/')
```

and links  looks like:

```
 dmc.NavLink(label="al", href="/al", active='partial'),
 dmc.NavLink(label="al2", href="/al/2", active='partial'),
```

Then if you click on the first link neither is active and when you click on the second link then only the `/al` page is active.

AnnMarieW: @BSd3v suggested this and it works well :slightly_smiling_face: 

```
   (active === 'partial' && (pathname.startsWith(href + '/') || pathname == href))

```

AnnMarieW: oh, I see that's what you had too @alexcjohnson :blush:   I was just fixated on the `href + '/'` part and missed it.

alexcjohnson (approved):
💃

PR #499, opened 2025-01-31
Updates for dash 3 

AnnMarieW:
Closes #453 

This PR replaces #458 which only handled changes for Dash 3.  This PR includes updates for  Dash 3 that are also compatible with Dash 2.

- [x] handle removal of loading_state
- [x] handle deprecation of defaultProps, including new way of setting persistence props
- [x] handle removal of `_dashprivate_loadingState`  used in the `Skeleton` component. 
- [x] update `Checkbox` component which has an icon component prop that needs to be rerender with additional props.
- [x] update all components that used `_dashprivate_layout`:   Checkbox, Popover,  Menu, Hovercard,  Timeline,  Stepper
- [x] Check PR  # 458 to make sure all changes are applied here too.
- [ ] In `package.json`, pin "dash-extensions-js": "0.0.8" . Before upgrading, check for Dash 2 compatibility. This package is required for the renderDashComponent function, which may be affected by updates in future Dash versions.  
- [x] Require dash >= 3.0 for install -- temporary change to  run tests using dash 3.  Reverted for pre-release.
- [ ] Set up tests to run for dash 2 and dash 3
-

**Note - do not use the new typing support because it does not handle Mantine typing correctly.**

ignore this comment in the build and add notes to releasing.md to ensure it's not added in the future

> !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
! Missing proptypes.js in `dash_mantine_components/__init__.py` !
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
>
>Add the following to `dash_mantine_components/__init__.py` to enable
runtime prop types validation with tsx components:

PR #506, opened 2025-02-06
Updates for Dash 3.0

### LINKED ISSUE #453: Update for Dash 3.0

AnnMarieW:
Replacement for PR 499.
**Please see #499 for more details and sample apps**

Closes #453 

This PR replaces #458 which only handled changes for Dash 3.  This PR includes updates for  Dash 3 that are also compatible with Dash 2.

- [x] handle removal of loading_state
- [x] handle deprecation of defaultProps, including new way of setting persistence props
- [x] handle removal of `_dashprivate_loadingState`  used in the `Skeleton` component. 
- [x] update `Checkbox` component which has an icon component prop that needs to be rerender with additional props.
- [x] update all components that used `_dashprivate_layout`:   Checkbox, Popover,  Menu, Hovercard,  Timeline,  Stepper
- [x] Check PR  # 458 to make sure all changes are applied here too.
- [x] In `package.json`, pin "dash-extensions-js": "0.0.8" . Before upgrading, check for Dash 2 compatibility. This package is required for the renderDashComponent function, which may be affected by updates in future Dash versions.  
- [x] Require dash >= 3.0 for install -- temporary change to  run tests using dash 3.  Reverted for pre-release.
- [ ] Set up tests to run for dash 2 and dash 3
-

**Note - do not use the new typing support because it does not handle Mantine typing correctly.**

ignore this comment in the build and add notes to releasing.md to ensure it's not added in the future

> !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
! Missing proptypes.js in `dash_mantine_components/__init__.py` !
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
>
>Add the following to `dash_mantine_components/__init__.py` to enable
runtime prop types validation with tsx components:


alexcjohnson (approved):
💃 Looks great. I'll note that there are a couple of places (`Checkbox` and a couple of charts are the ones I noticed) where you omitted a default value of `false` that was previously explicit. This shouldn't change the behavior at all, it'll just change the docstring, but to my eye leaving these implicit is intuitive.

PR #521, opened 2025-03-02
async loading

### LINKED ISSUE #367: Large Bundle Size for dash_mantine_components.js
The current `dash_mantine_components.js` file is quite large at 2.7MB, which can impact performance and load times.

### Build Output:
```

> dash_mantine_components@0.14.6 build:js
> webpack

asset dash_mantine_components.js 2.6 MiB [compared for emit] [minimized] [big] (name: main) 1 related asset
orphan modules 1.59 MiB [orphan] 726 modules
runtime modules 1.29 KiB 7 modules
modules by path ./node_modules/ 5.18 MiB 1588 modules
modules by path ./src/ts/ 321 KiB
  modules by path ./src/ts/components/core/ 239 KiB 127 modules
  modules by path ./src/ts/components/charts/*.tsx 15.3 KiB 8 modules
  modules by path ./src/ts/components/extensions/ 14.1 KiB 8 modules
  modules by path ./src/ts/components/dates/*.tsx 24.9 KiB 7 modules
  modules by path ./src/ts/utils/*.ts 5.24 KiB
    ./src/ts/utils/charts.ts 655 bytes [built] [code generated]
    + 3 modules
  ./src/ts/index.ts 21.2 KiB [built] [code generated]
  ./src/ts/components/styles/MantineProvider.tsx 1.49 KiB [built] [code generated]
external {"commonjs":"react","commonjs2":"react","amd":"react","umd":"react","...(truncated) 42 bytes [built] [code generated]
external {"commonjs":"react-dom","commonjs2":"react-dom","amd":"react-dom","umd"...(truncated) 42 bytes [built] [code generated]

WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  dash_mantine_components.js (2.6 MiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  main (2.6 MiB)
      dash_mantine_components.js


WARNING in webpack performance recommendations: 
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/




```
### Questions:

 **Orphan Modules:**
- Are the orphan modules necessary for the current dependencies?
- If not, can they be removed?
- What is the best way to identify and eliminate unnecessary orphan modules?

**Lazy Loading Components:**

- What is the most efficient way to determine which components should be lazy-loaded?
- A good starting point may be components like charts, dates, and the carousel component. Would these be good candidates for lazy loading, or are there others to consider?

**Additional Strategies to Reduce Bundle Size:**
- Are there any other techniques or optimizations we could apply?

@alexcjohnson - I would love your input on this topic

AnnMarieW:
Closes #367 

### Implementing Async Loading for Improved Performance

This  will reduce initial load times and improve performance, especially for apps using only a subset of the available components.

Reduction in `dash_mantine_components.js` file size:
-  before changes: 2.68 MiB
- after async loading for all the charts:  1.71MiB :tada:
- after async loading for all code highlight components: 823 KiB :partying_face: 

------------

Maybe move the following to a future PR?

Add async loading to all the files in these folders:
 - dates
 - notifictions
 - npropgress
 - core:
    

In the core folder, the largest files only:
Size (KB) | File Name
-- | --
56 | MultiSelect.py
52 | TagsInput.py
52 | Select.py
48 | Autocomplete.py
44 | Modal.py
40 | ColorInput.py
36 | LoadingOverlay.py
32 | Tooltip.py
32 | Drawer.py
28 | NumberInput.py
28 | JsonInput.py
28 | FloatingTooltip.py
28 | Button.py
28 | ActionIcon.py
24 | TimeInput.py
24 | TextInput.py
24 | Textarea.py
24 | RangeSlider.py
24 | PasswordInput.py


AnnMarieW:
@alexcjohnson

I'm just getting started on this project and have applied async loading to the `AreaChart` and `BarChart`. With just these two components updated, the **`dash_mantine_components.js`** file size has already been reduced from **2.68 MiB to 2.18 MiB** 🎉.  

I'm following the approach Emil used for async loading in `dash-leaflet`:  
- Example of a component with async loading: [EditControl.tsx](https://github.com/emilhe/dash-leaflet/blob/master/src/ts/components/EditControl.tsx)
- This commit helped me set up async loading: [Async Loading Commit](https://github.com/emilhe/dash-leaflet/commit/788b329f4b725a7b471c4832144d494ad0e50433)

Before applying these changes to all the other files, I’d love to get your input on the following:  

- Is it okay to place a `fragments` folder inside each component directory (`charts`, `core`, `dates`, `extensions`)? Any potential issues with this approach?  
- There seems to be some redundancy in prop definitions between the component and the fragment. Is there a way to optimize this?  
-  Am I missing anything important, or is there anything unnecessary in the current setup?  
- Any other comments?

Looking forward to your thoughts! 😊

alexcjohnson:
> Is it okay to place a fragments folder inside each component directory (charts, core, dates, extensions)? Any potential issues with this approach?

That should be fine.

> There seems to be some redundancy in prop definitions between the component and the fragment. Is there a way to optimize this?

You can't directly import the fragment in the wrapper, but the reverse is OK: you should be able to import the interface from the wrapper, like we do with props in [some older components](https://github.com/plotly/dash/blob/a00407dcd0779e01372ec14c21bb4f19f11df56b/components/dash-core-components/src/fragments/DatePickerRange.react.js#L6)

> In the core folder, the largest files only

These are the Python files? Those don't say much about the size of the code they use, just about the props they support but that's heavily duplicated vs the JS code. I'd go one-by-one thinking about the complexity of what these components likely do in JS, and make fragments according to your judgment on that. So great call starting with charts, after which datepickers seem like a great next step but I don't yet have a great sense of what all is in Mantine to know if there are other candidates for high-code components.

AnnMarieW:
@alexcjohnson 

What do you think  is the minimum size difference to make it worthwhile doing the async loading?  I just tried the `Carousel` and `CarouselSlide` and I was surprised to see it didn't make that much difference.  The `dash_mantine_components.js` went from 1.71 to 1.69 MiB

emilhe:
> @alexcjohnson
> 
> What do you think is the minimum size difference to make it worthwhile doing the async loading? I just tried the `Carousel` and `CarouselSlide` and I was surprised to see it didn't make that much difference. The `dash_mantine_components.js` went from 1.71 to 1.69 MiB
> 
> Next did the three code highlight components: `CodeHighlight`, `CodeHighlightTabs` `InlineCodeHighlight` - those made a big difference!
> 
> * `dash_mantine_components.js` (823 KiB)
> * `dash_mantine_components-shared.js` (1.33MiB)
> 
> However, now that the `dash_mantine_components-shared.js` is so large, is there still the same benefit? If people don't use the charts or the code highlight components, will this shared file not have to be loaded?

I have made a [PR trying to mitigate this issue](https://github.com/AnnMarieW/dash-mantine-components/pull/2).

alexcjohnson (approved):
💃 Great job @AnnMarieW - and fantastic contribution @emilhe 🎉

AnnMarieW:
Thanks @emilhe :tada:

Issue #522, opened 2025-03-03
Notifications show unexpectedly


AnnMarieW:
It was reported on the DMC discord that Notification would show randomly.

Here is a sample app - the notification will show after the app is canceled.

```python
import dash_mantine_components as dmc
from dash import Dash, _dash_renderer, callback, Output, Input
from dash_iconify import DashIconify

_dash_renderer._set_react_version("18.2.0")
app = Dash(external_stylesheets=dmc.styles.ALL)
 
app.layout = dmc.MantineProvider(
    children = [
        dmc.NotificationProvider(),
        dmc.Affix(id = 'notify-container', position={"bottom": 20, "right": 20}),
        dmc.Button("Load Data", id="show-notification")
    ],
)

@callback(
    Output("notify-container", "children"),
    Input("show-notification", "n_clicks"),
    prevent_initial_call=True,
)
def notify(n):
    nf = dmc.Notification(
                    id="my-notification",
                    title="Appointment booked",
                    message="Your appointment has been schedueled successfully",
                    color="green",
                    loading=False,
                    action="show",
                    autoClose=2000,
                    icon=DashIconify(icon="akar-icons:circle-check"),
                )
    return nf

if __name__ == "__main__":
    app.run(debug=True)


```

AnnMarieW:
It is possible to clear notifications using `action="clean"` or `action="cleanQueue"`  Not sure how to make use of it in a real app.  In this app, if you hit the "clean" button before canceling the app, you don't see the extra notification pop up after the app is cancelled.

```python

import dash_mantine_components as dmc
from dash import Dash, _dash_renderer, callback, Output, Input, ctx
from dash_iconify import DashIconify

_dash_renderer._set_react_version("18.2.0")
app = Dash(external_stylesheets=dmc.styles.ALL)

app.layout = dmc.MantineProvider(
    children=[
        dmc.NotificationProvider(position="bottom-left"),
        dmc.Affix(id='notify-container', position={"bottom": 20, "right": 20}),
        dmc.Button("Load Data", id="show-notification"),
        dmc.Button("clean", id="clean-notification")
    ],
)


@callback(
    Output("notify-container", "children"),
    Input("show-notification", "n_clicks"),
    Input("clean-notification", "n_clicks"),
    prevent_initial_call=True,
)
def notify(n,n2):
    if ctx.triggered_id == "show-notification":
        return dmc.Notification(
            id="my-notification",
            title="Appointment booked",
[...]

PR #523, opened 2025-03-03
adding the ability for notifications to auto hide when they close

BSd3v:
keeps from showing up unexpectedly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 # [Unreleased]
 
 ### Changed
-- `notification` will not automatically set its `action` to `hide` when closed, this avoids issues where a `callback` error would re-trigger the component.
+- `notification` will now automatically set its `action` to `hide` when closed, this avoids issues where a `callback` error would re-trigger the component.
 
 # 1.0.0rc2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Change Log
 
+# [Unreleased]
+
+### Changed
+- `notification` will not automatically set its `action` to `hide` when closed, this avoids issues where a `callback` error would re-trigger the component.
+
 # 1.0.0rc2
 
 ### Added
--- a/src/ts/components/extensions/notifications/Notification.tsx
+++ b/src/ts/components/extensions/notifications/Notification.tsx
@@ -41,6 +41,10 @@ interface Props extends BoxProps, StylesApiProps, Omit<DashBaseProps, "id"> {
 const Notification = (props: Props) => {
     const { action, setProps, loading_state, ...others } = props;
 
+    others['onClose'] = () => {
+        setProps({'action': 'hide'})
+        }
+
     useEffect(() => {
         switch (action) {
             case "show":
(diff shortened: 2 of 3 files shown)


src/ts/components/extensions/notifications/Notification.tsx
```diff
@@ -41,18 +41,18 @@ interface Props extends BoxProps, StylesApiProps, Omit<DashBaseProps, "id"> {
...
+    const onClose = () => {
         setProps({'action': 'hide'})
-        }
+    }
 
     useEffect(() => {
         switch (action) {
             case "show":
-                notifications.show(others);
+                notifications.show({...others, onClose});
```
AnnMarieW, line 51: how about making it `{onClose, ...others}` incase we get the functions as props working in a future version?

BSd3v: At that point, we will need to rework this anyways. As you wont be able to declare the name twice, plus this default behaviour should stay unless explicitly dont not to auto remove from the tree... At which time, this would trigger the event passed to the props while maintaining the original removal.

AnnMarieW:
Closes #522 

Thanks for the PR @BSd3v   I ran it with the sample app in issue 522 and it solves the problem.  What do you think about adding a test?  Also need a changelog entry

BSd3v:
@alexcjohnson,

What are your thoughts on adding a prop for a dev to be able to opt out of this? Where the prop would stay as they added without it auto `hide` for the action.

Not sure why the dev would want this behaviour though...

alexcjohnson:
This does not need an opt-out. If it did, to be consistent we should in fact re-show the notification immediately when it's closed, because internally it still thinks its action is `show` even though it has been closed. The root of the issue is we're operating this halfway between a controlled and an uncontrolled component, but your fix makes it closer to a fully-controlled component (even if it's redundant control because the close has already happened).

The only thing I'm wondering is if `hide` is really the right new action, or if it should be `none` or something? I guess what this is pointing to is this component is written in a weird way - `action` isn't really a prop, it's, well, an action. It's a change of state, but given that React expects to be able to rerender to its heart's content without anything changing if it's given consistent props, it doesn't really work to have a prop that represents a change. Ideally the prop should just be "is this notification visible or not", then the component or the notification system figures out when it sees some props, does this represent a new component to show, a currently-visible component I need to hide, or an existing component I need to update.

Anyway that's probably a bigger task than you want to do right now... it must already work (ie be a no-op) to re-show an already-visible notification or you'd have seen that bug pop up way sooner, so this seems like it'll work if people use it in simple ways, but there are very likely other bugs hiding out in more complex use cases.

So I'd say go for it, merge this as a fix but down the line we can think about a cleaner pattern.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+# [Unreleased]
+
+### Changed
+- `notification` will not automatically set its `action` to `hide` when closed, this avoids issues where a `callback` error would re-trigger the component.
```
alexcjohnson, line 6:
```suggestion
- `notification` will now automatically set its `action` to `hide` when closed, this avoids issues where a `callback` error would re-trigger the component.
```

alexcjohnson (approved):

BSd3v:
Believe this issue stems from treating an underlying `api` as a component which is then rerendered by `dash`.

I recommend that we shift to treating the `notifications` as new prop of the `NotificationProvider` and treat it as an `api` call similar to how `rowTransactions` work with `dash-ag-grid`. This should keep the rerendering of the components and the notification system the way that it was designed by `Mantine`.

Issue #480, opened 2025-01-15
[Feature Request] Add a function to generate Mantine custom colors


AnnMarieW:
It looks like the [Mantine Colors Generator](https://mantine.dev/colors-generator/) is available as a separate package: https://www.npmjs.com/package/@mantine/colors-generator

It would be nice to be able to generate a custom color sequence from within an app.  

```
import { generateColorsMap } from '@mantine/colors-generator';

const hexString = "#FFFFFF";
const { colors, baseColorIndex } = generateColorsMap(hexString);

// colors = MantineColorsTuple
// baseColorIndex = MantineTheme.primaryShade index

```

AnnMarieW:
I'm not sure if it's worthwhile adding this feature.  In the [Mantine 7.0 Changelog](https://mantine.dev/changelog/7-0-0/#colors-generator) they note that:
> It is also available as [online tool](https://mantine.dev/colors-generator/). Note that it is usually better to generate colors in advance to avoid contrast issues.

I'm going to close this for now, but it would be fairly easy to add.  I already have a PR:  https://github.com/snehilvj/dash-mantine-components/pull/527

If you would like this feature, please add a comment there and we can reconsider adding this in the future if there is enough interest.

deadkex:
@AnnMarieW Why not add it as python function/class/x? That way the bundle size wouldnt increase but we still get the functionality.
Converted with ChatGPT: https://chatgpt.com/share/67cfff09-1420-8009-af58-1154d77b3df0

PR #527, opened 2025-03-10
Added ability to generate list of 10 colors in the theme

AnnMarieW:
closes #488 

Adds a `colorsGenerator` prop to the `MantineProvider`

 Dictionary of colors, where each key is a color name and the value is a base color string.  Each base color is passed to the Mantine `generateColors` function, which returns an array of 10 color shades.
     
Example:
      `colorsGenerator = { "light-blue": "#375EAC" }`
      
Will be transformed into:
      `{ "light-blue": ["#eff5fc", "#d6e5fa", ..., "#162d50"] }`

This is merged with the `theme.color` object

The `generateColors` function was added to Mantine in V7.0

> New [@mantine/colors-generator](https://mantine.dev/theming/colors#colors-generation) package is now available to generate
color palettes based on single color value. It is also available as [online tool](https://mantine.dev/colors-generator/).
Note that it is usually better to generate colors in advance to avoid contrast issues.

Question:
- given that it's recommended to use the on-line tool, is it worthwhile adding?  It does increase the bundle size some (about 60 KiB)

----------

Here's a sample app:
```

from dash import Dash, _dash_renderer, Input, Output, State

    dmc.Alert("hello dmc"),
    colorsGenerator={'pale-blue': '#375EAC', "another-red": "#f0185c"},
    theme={"primaryColor": "another-red"}
)

if __name__ == "__main__":
    app.run(debug=True)




```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,7 @@
                 "@mantine/carousel": "7.17.0",
                 "@mantine/charts": "7.17.0",
                 "@mantine/code-highlight": "7.17.0",
+                "@mantine/colors-generator": "7.17.0",
                 "@mantine/core": "7.17.0",
                 "@mantine/dates": "7.17.0",
                 "@mantine/hooks": "7.17.0",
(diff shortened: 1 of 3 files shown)


AnnMarieW:
May not be worthwhile adding.  See Issue #480 for more details.  Closing for now and may re-open some time in the future.

PR #530, opened 2025-03-13
Add RichTextEditor component

emilhe:
This PR adds a new `RichTextEditor`, a wrapping of the [most recent RTE componnent from Mantine](https://mantine.dev/x/tiptap/) based on tiptap.  [Original PR](https://github.com/AnnMarieW/dash-mantine-components/pull/1)

### Extensions

The tiptap component comes with a huge number of [extensions](https://tiptap.dev/docs/editor/extensions/overview). In the current design, support is hardcoded for a number of predefined extensions, which can then be selectively enabled by passing a list of strings (or dict, if you need to provide options) from the Python layer with the names of the plugins to enable. Note however that all supported plugins are included in the component chunk (!). I would love to hear if anyone has ideas for a better design (the currently added plugins are small though, so at present bundle size increase due to plugins is not a huge issue), and/or which plugins should be bundled.

### Toolbar

The toolbar construction is performed in React using a ReactNode hierarchy (i.e. components). My initial idea was to pass a such hierarchy from Dash direcly (to keep the design as aligned with Mantine as possible), but as there are so many controls, that would require the creation > 10 components just for toolbar configuration. As an alternative, I have created a design, where a JSON-like structure with the component names  (or dict, if you need to provide options) is passed instead (similar to how the plugins are passed). 

### Content

The content can be provided as either HTML via the `html` property or as ProseMirror JSON via the `json` property. Both properties are controlled, i.e. when data in the editor changes, callbacks listening to `html` and/or `json` are triggered. Debounce is supported similar to other DMC components. The rendering of the content as HTML introduces a slight overhead for large texts, so the API may be extended in the future with an option to disable HTML rendering. For now, performance tuning is limited to the debounce property.

### Selection

The currently selected text can be read from Dash via the `selected` property. Its affected by the debounce property, similar to the  `html` /  `json`  properties. The returned data does *not* include any formatting.

###  Bundle size

The `RichTextEditor` component supports async loading. With the current design, the increase in the (main) bundle is therefore small (about 3kB, which I believe is acceptable). The component itself (including currently added plugins) yields a chunk > 300 kB size. As more plugins are added, the size will increase. At some point, we may consider splitting the plugins into chunks.

## Example(s)

The following (Dash 3) example is a is inspired by the first/main example in [the Mantine docs](https://mantine.dev/x/tiptap/), but with some additional extensions/controls added,

```
from dash import Dash, Input, html

# Content from the Mantine example.
content = """<h2 style="text-align: center;">Welcome to Mantine rich text editor</h2><p><code>RichTextEditor</code> component focuses on usability and is designed to be as simple as possible to bring a familiar editing experience to regular users. <code>RichTextEditor</code> is based on <a href="https://tiptap.dev/" rel="noopener noreferrer" target="_blank">Tiptap.dev</a> and supports all of its features:</p><ul><li>General text formatting: <strong>bold</strong>, <em>italic</em>, <u>underline</u>, <s>strike-through</s> </li><li>Headings (h1-h6)</li><li>Sub and super scripts (<sup>&lt;sup /&gt;</sup> and <sub>&lt;sub /&gt;</sub> tags)</li><li>Ordered and bullet lists</li><li>Text align&nbsp;</li><li>And all <a href="https://tiptap.dev/extensions" target="_blank" rel="noopener noreferrer">other extensions</a></li></ul>"""

# Colors from the Mantine example.
colors = [
    "#25262b",
    "#868e96",
    "#fa5252",
    "#e64980",
    "#be4bdb",
    "#7950f2",
    "#4c6ef5",
    "#228be6",
    "#15aabf",
    "#12b886",
    "#40c057",
    "#82c91e",
    "#fab005",
    "#fd7e14",
]

# Setup the Dash app.
app = Dash(__name__, external_stylesheets=dmc.styles.ALL)
    html.Div(
        dmc.RichTextEditor(
            html=content,
            labels={
                "boldControlLabel": "Bold. Wuhu!",
                "colorControlLabel": "This is a {color} color. Wuhu!",
            },
            toolbar={
                "sticky": True,
                "controlsGroups": [
                    [
                        "Bold",
                        "Italic",
                        "Underline",
                        "Strikethrough",
                        "ClearFormatting",
                        "Highlight",
                        "Code",
                    ],
                    ["H1", "H2", "H3", "H4"],
                    [
                        "Blockquote",
                        "Hr",
                        "BulletList",
                        "OrderedList",
                        "Subscript",
                        "Superscript",
                    ],
                    ["Link", "Unlink"],
                    ["AlignLeft", "AlignCenter", "AlignJustify", "AlignRight"],
                    ["Undo", "Redo"],
                    [
                        {"Color": {"color": "red"}},
                        {"Color": {"color": "green"}},
                        {"Color": {"color": "blue"}},
                    ],
                    [
                        {"ColorPicker": {"colors": colors}},
                        "UnsetColor",
                    ],
                ],
            },
            id="rte",
        ),
        style=dict(margin="5px"),
    )
)


# Read the selected text from the editor.
@app.callback(Input("rte", "selected"))
def update_content(content: str):
    print(content)


if __name__ == "__main__":
    app.run(debug=True)
```


AnnMarieW:
The `n_blur` is an additional prop for when the debounce is set to True.  The `n_blur` can also be used to trigger a callback.

Here is how I defined `debounce` in my example - it's consistent with how `debounce` works in Dash and other DMC components:

> (boolean | number; default False): If True, changes to input will be sent back to the Dash server only on enter or when losing focus. If it's False, it will send the value back on every change. If a number, it will not send anything back to the Dash server until the user has stopped typing for that number of milliseconds.

---------------


I think it would be a very strange pattern to set content in one prop and track changes in another.  I can't think of anywhere else in Dash where it's done this way.

I liked the original `format` prop.  The only thing I'd change is not to allow undefined (untracked)  It would be hard to debug if it was unexpectedly set as something other than html or json.


>  if you choose json as format and input html, the content property "will change itself" (i.e. just after setting it, it will not be the same value as you inputted, which is a bit weird)

Actually, this might be a feature instead of bug?  Perhaps this would be a way to convert from one format to another.

All that being said, I did play around with your version 3, and it worked quite well.   I'd like to hear @alexcjohnson 's thoughts on this too - he's so good with Dash API design. 

​

alexcjohnson:
> I think it would be a very strange pattern to set content in one prop and track changes in another. I can't think of anywhere else in Dash where it's done this way.

Yeah let's see if we can avoid that. I wonder if we could do `html` and `json`, either of which would accept input and provide output? We'd have to decide on precedence if you provided both initially, but then as changes come in you just accept whichever one changed and use it to update the other? Also, do we really need `track_html` and `track_json`? Unless it's really slow for the editor to calculate these, there should be very little overhead to just setting those props if you don't have a callback listening to them. I'm imagining something like:
```js
useDidUpdate(() => {
    if (editor && editor.getHTML() !== html) {
        editor.commands.setContent(html);
    }
    syncDashState();
}, [html]);

useDidUpdate(() => {
    // This test may need to be stringified if editor.getJSON makes a new object?
    if (editor && editor.getJSON() !== json) {
        editor.commands.setContent(json);
    }
    syncDashState();
}, [json]);

const syncDashState = () => {
    const update = {}
    if (editor) {
        const newHTML = editor.getHTML();
        const newJSON = editor.getJSON();
        let hasUpdate = false;
        if (html !== newHTML) {
            update.html = newHTML;
            hasUpdate = true;
        }
        // This test may need to be stringified if editor.getJSON makes a new object?
        if (json !== newJSON) {
            update.json = newJSON;
            hasUpdate = true;
        }
        if (hasUpdate) {
            setProps(update);  // even if both changed, we only call setProps once
        }
    }
}
```

emilhe:
Since json is the native format, I think it would make most sense that it takes precedence. In terms of performance, the `getJSON` call seems to be very fast, about ~ 0.1 ms even for large texts. So I guess it would be an option always to track json. The `getHTML`  call is slower, and it seems to take longer for longer texts. I am thus more hesitant to enable the html tracking without allowing the user to opt out.

BSd3v:
I'd almost recommend defaulting to just providing data back as JSON, and then maybe expose the api a bit in order to fetch the HTML on demand? Could have a prop to fetch the state to place into a different prop rather than content/value.

To me, it makes more sense to store in JSON in a db and then export to something like emails in HTML.

Since this is an all in one, essentially, couldnt we add a dash button for submitting the html to export.

emilhe:
I have just completed an implemention of the design suggested by @alexcjohnson . At a glance, I think I like it,

(+) when you set a prop called html/json it's very clear what type of input data is expected
(+) same goes for the case when the prop is callbacks; it's clear what type of data is returned from the component
(+) types are well defined, json is object and html is string (both on the input/output side)
(+) the api is simple, only two props to "comprehend", html and json
(-) performance may be an issue for large documents (due to the `getHTML` call). That being said, in the current design this call is "protected" by the debounce, so the user can mitigate this issue (at least partially) but increasing the debounce. Finally, if we decide to add a `track_html` property in the future (if it turns out that performance does become an issue), we can do so with a non-breaking change as long as we keep the default value to `true`

I'll look more into the blur stuff tomorrow between meetings.

emilhe:
I have just added a new `selected_text` property that tracks the currently selected text (without formatting). In the initial implementation, I have used the same `debounce` property to control the update rate as for `html`/`json`. I also considered adding another proerty, e.g. `debounce_selected_text` to enable more fine grained control. On the other hand, using the same prop yields a simpler API design. I was also a bit in double about the prop name; should it just be `selected`? Or `selected_text`? Or ... I have put `selected_text` for now to emphase that what is returned is *pure text* (no formatting/structure), in contrast to the `html`/`json` that return a particular format.

src/ts/components/extensions/richtexteditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,57 @@
...
+type Extension = "StarterKit" | "Underline" | "Underline" | "Link" | "Superscript" | "SubScript" | "Highlight" | "TextAlign" | "Table" | "TableCell" | "TableRow" | "TableHeader";
+
+/** Supported controls (add more as needed). */
+type Control = "Bold" | "Italic" | "Underline" | "Strikethrough" | "ClearFormatting" | "Highlight" | "Code" | "H1" | "H2" | "H3" | "H4" | "Blockquote" | "Hr" | "BulletList" | "OrderedList" | "Subscript" | "Superscript" | "Link" | "Unlink" | "AlignLeft" | "AlignCenter" | "AlignJustify" | "AlignRight" | "Undo" | "Redo";
+
+/** Toolbar property definition. */
+type Toolbar = {
+  sticky?: boolean;
+  controlsGroups?: Control[][];
+}
```
AnnMarieW, line 19: Would it be better to define these types in the props folder?

AnnMarieW:
@emilhe 
Can you also add the missing props?  https://mantine.dev/x/tiptap/?t=props

AnnMarieW:
This question comes up frequently on the Mantine Discord:

Looks like we will need to add the [TypographyStylesProvider](https://mantine.dev/core/typography-styles-provider/) component.  But it's OK to just open an issue and it can be handled in a separate PR.

image

--------------

image

src/ts/components/extensions/richtexteditor/RichTextEditor.tsx
emilhe: As a general rule, I would typically hold models definitions (here, props) close to their usage (i.e. here) when they are only used in a single place, and only move them when they start to be used in multiple places (like e.g. `DashBaseProps`). That being said, I am happy to follow any style guides for the library. Would you like them moved?

emilhe:
I have just renamed the selection property from `selected_text` to `selected`. Furthermore, I have extended the API to support passing options to extensions. I have also added this missing props. I have added `withTypographyStyles` for now to ensure alignment with the Mantine API, although it may not really make sense until the `TypographyStylesProvider` component is available. I would prefer to add that component in a separate PR though to limit the scope of this PR.

AnnMarieW:
Thanks for adding the `TypographyStylesProvider` :rocket: :


On another note, is it odd that the capitalization is inconsistent? `Subscript` and `SubScript`, but it's always `Superscript` and not `SuperScript`

This is from the  [Mantine docs:](https://mantine.dev/x/tiptap/#usage) so not much we can/should do about it?
```
import Superscript from '@tiptap/extension-superscript';
import SubScript from '@tiptap/extension-subscript';
```
```
          <RichTextEditor.Subscript />
          <RichTextEditor.Superscript />
```

emilhe:
> On another note, is it odd that the capitalization is inconsistent? `Subscript` and `SubScript`, but it's always `Superscript` and not `SuperScript`

I think that's just a typo that I carried over from Mantine (it doesn't seem to be present in tiptap). I just corrected it (to lowercase), and it seems to work just fine. I have also added the placeholder, color, and text style plugins.

AnnMarieW:
> other extensions that you think we should include in the initial release?

I think this is a good place to start - we can always add more later.

The `dash_mantine_components-shared.js` file is now up to 1.4 MB.  Most of that is the code highlight components.  What do you think about moving those to a separate file like how the charts are handled?

emilhe:
> The `dash_mantine_components-shared.js` file is now up to 1.4 MB. Most of that is the code highlight components. What do you think about moving those to a separate file like how the charts are handled?

What branch are you one? I don't see a `dash_mantine_components-shared.js` block at all,

image

If you are referring to a file, could it be legacy from an old build before we changes the chunking logic?

AnnMarieW:
Yes, you are correct - it was from an old build.  Please disregard. :see_no_evil:

CHANGELOG.md
```diff
@@ -1,65 +1,76 @@
 # Change Log
 
+# 1.0.1
```
AnnMarieW, line 3: We may call this 1.1.0, but for now, let's just keep it as
```suggestion
# [UNRELEASED]
```

emilhe: Good point. I guess it may actually end with 1.1.0, as we are adding (significant) functionality.

AnnMarieW:
For the labels, the `colorPickerColorLabel` and the `colorControlLabel` are functions.  For example, this is the default:
```
 colorControlLabel: (color) => `Set text color ${color}`,
colorPickerColorLabel: (color) => `Set text color ${color}`,
```

I expect the main purpose for setting the labels prop is for localization.  Do you think there is a way to allow a string to update the "Set text color" string in the function?  

If not, I could just mark this as a limitation in the docs.

https://mantine.dev/x/tiptap/#labels-and-localization

emilhe:
> For the labels, the `colorPickerColorLabel` and the `colorControlLabel` are functions. For example, this is the default:
> 
> ```
>  colorControlLabel: (color) => `Set text color ${color}`,
> colorPickerColorLabel: (color) => `Set text color ${color}`,
> ```
> 
> I expect the main purpose for setting the labels prop is for localization. Do you think there is a way to allow a string to update the "Set text color" string in the function?
> 
> If not, I could just mark this as a limitation in the docs.
> 
> https://mantine.dev/x/tiptap/#labels-and-localization

Ah, yes. There are a few functions. I guess there are a number of ways to address it. Are you aware of any precedence in Mantine tha we should follow? Otherwise, one possibly option would be to use f-string syntax? That's simple and Pythonic.

AnnMarieW:
> one possibly option would be to use f-string syntax? That's simple and Pythonic.

True, but I think it needs to be a javascript function.  We haven't enabled passing functions as props yet.

src/ts/components/extensions/richtexteditor/RichTextEditor.tsx
```diff
@@ -0,0 +1,96 @@
...
+  | "Image"
+  | "Color";
+
+// TODO: Maybe add types for the extensions options explicitly?
+type Extension =
+  | ExtensionName
+  | { [key in ExtensionName]?: Record<string, unknown> }; // Allow any options
+
+/** Supported controls (add more as needed). */
+type ControlName = "Bold" | "Italic" | "Underline" | "Strikethrough" | "ClearFormatting" | "Highlight" | "Code" | "H1" | "H2" | "H3" | "H4" | "Blockquote" | "Hr" | "BulletList" | "OrderedList" | "Subscript" | "Superscript" | "Link" | "Unlink" | "AlignLeft" | "AlignCenter" | "AlignJustify" | "AlignRight" | "Undo" | "Redo" | "Color" | "ColorPicker" | "UnsetColor";
```
AnnMarieW, line 34: missing H5, H6, CodeBlock

emilhe:
> > one possibly option would be to use f-string syntax? That's simple and Pythonic.
> 
> True, but I think it needs to be a javascript function. We haven't enabled passing functions as props yet.

In principle, yest. But if we are only doing f-string, that can be mapped with a (relatively) simple replacement. Here is an example implementation,

https://github.com/emilhe/dash-mantine-components/pull/1/files

that makes is possible to pass labels like this,

```
            labels={
                "boldControlLabel": "Bold. Wuhu!",
                "colorControlLabel": "This is a {color} color. Wuhu!",
            },
```

image

My main concern on adding it is that we form precedence on how to pass functions in dmc. And that maybe a topic that needs additional discussion. That being said, for the task at hand, the implementation proposal works nicely.

src/ts/components/extensions/richtexteditor/RichTextEditor.tsx
emilhe: Nicely spotted! Thanks! They should be there now

AnnMarieW:
> My main concern on adding it is that we form precedence on how to pass functions in dmc. And that maybe a topic that needs additional discussion. 

You make a great point—I share the same concern. Maybe @alexcjohnson can weigh in on this as well.

Allowing functions as props is a high priority for me, but I’m not sure how long it will take to implement. If there’s a  long delay, one option is to document the limitation and implement your solution if it's a popular feature request.


> That being said, for the task at hand, the implementation proposal works nicely.

I agree!  I especially like the ability to add "Wuhu!"  :laughing:

AnnMarieW:
It would be good to add persistence.  Here's an example - with the [Switch](https://github.com/snehilvj/dash-mantine-components/blob/b6f1ba4a8cdeccda3ddac26be909ad491237247d/src/ts/components/core/Switch.tsx) component

Will probably need to make both `json` and `html` persistent

AnnMarieW:
I've been working on the docs:  https://github.com/snehilvj/dmc-docs/pull/178

Still a little more content to add and some final edits, but it's coming along nicely.

This is my new favorite component!  Thanks for adding it @emilhe

src/ts/components/extensions/richtexteditor/fragments/RichTextEditor.tsx
```diff
@@ -0,0 +1,185 @@
...
+import Superscript from "@tiptap/extension-superscript";
+import Subscript from "@tiptap/extension-subscript";
+import Table from "@tiptap/extension-table";
+import TableCell from "@tiptap/extension-table-cell";
+import TableHeader from "@tiptap/extension-table-header";
+import TableRow from "@tiptap/extension-table-row";
+import Placeholder from "@tiptap/extension-placeholder";
+import Color from "@tiptap/extension-color";
+import TextStyle from '@tiptap/extension-text-style';
+import Image from '@tiptap/extension-image';
```
AnnMarieW, line 20: Since all these extensions need to be imported in the component, is there any reason not to enable them all by default?  Is there a performance difference?

AnnMarieW: All good points.  Let's leave it as-is. :heavy_check_mark:

emilhe:
> It would be good to add persistence. Here's an example - with the [Switch](https://github.com/snehilvj/dash-mantine-components/blob/b6f1ba4a8cdeccda3ddac26be909ad491237247d/src/ts/components/core/Switch.tsx) component
> 
> Will probably need to make both `json` and `html` persistent

I have added persistance is [this commit](https://github.com/snehilvj/dash-mantine-components/pull/530/commits/446642680983af2d854777360de03cf905c5fa44). I haven't done it before - but it seems to work; it seemed almost too simple?

emilhe:
> @emilhe - This looks great!  Any outstanding issues (besides the functions in the labels)?  If not, I think it's ready for a final review by Alex.

The only things if have noted are,

1) Your suggestion on adding all extensions by default. I am still trying to decide which is the best approach 🙃

2) If the f-string syntax for colors should be merged or not

Maybe Alex has an opinion on one or the other 😃

AnnMarieW:
If we decide not to include all the extensions by default, I made it easy to copy all of them from the docs:

image

alexcjohnson:
re: label functions - these ones can be cast as more "templates" than functions, in which case we don't really set a function precedent. I'm fine with f-string notation, I'll note that plotly.js / plotly.py has its own [template format](https://plotly.com/python/hover-text-and-formatting/#customizing-hover-text-with-a-hovertemplate) that uses the syntax `%{variable}`. But I doubt many people know this syntax, and it's a bit annoying that this is neither the Python nor the JS template syntax (IIRC we created this before JS templates were a thing, and before Python f-strings though `str.format()` would have existed) so probably not worth copying.

re: persistence, yes it's indeed simple, dash-renderer does all the work here 😎. This is a bit of a funny case though, in principle `html` and `json` are redundant, but also I wonder if there's a bug you can generate because of how they interact, something like:
- the app sets the `html` prop only, with some non-empty value
- the component translates it into the `json` prop.
- then the user types in their own text
- it gets translated to both `html` and `json` and stored in persistence
- reload the app and have it set a DIFFERENT non-empty `html` prop
- this SHOULD use the new `html` prop to generate the text in the component rather than the persisted value.
- but because `json` takes precedence - and as far as the component is concerned I think the initial value of `json` it will compare to is empty, which is what it sees the second time too - it will load the `json` value from persistence and that's what you'll get.

Play with this a bit, see if there is indeed a bug here. If there is, I'm not seeing an obvious fix, we may just need to document that if you want both a variable initial value and persistence you should set `persisted_props` to just the prop you intend to use to pass in the initial value, ie `['html']` in this example.

emilhe:
re: label functions - that's a good pointe. With the plotly template syntax not being widely adopted, I don't think that's the right way to go. I am leaning more towards fstring, which seems to be the most used syntax in Python today, and also available in R and Julia (through packages). Alternatively, we should adopt JS template syntax, i.e. `${color}`. That aligns will with the fact that the string is actually evaluated in JS, with the downside being that the syntax is probably less fimilar to most Dash users.

emilhe:
I kept getting test failures. I have added a commit to fix it here, but also opened a [separate PR](https://github.com/snehilvj/dash-mantine-components/pull/538) (that seems more appropriate, as the failing test is not really related to this work).

AnnMarieW:
OK, I see you added all the extensions by default. 
 
So if you want to change the default, like the text for the placeholder, it will be necessary to include all the extensions to keep the default behavior.    Do you think people will find that confusing?  I can update the docs to emphasis that but want to hear your thoughts first

It would be nice to just merge and changes to the extensions into the default, but then there would be no way to exclude an extension.  Not sure why people would want to exclude any, but one that comes to mind might be links.

emilhe:
@AnnMarieW Good points. While it would be nice to only update "what needs to be updated", I don't like the idea of not being able to disable extensions. I could imagine cases where you would want not to include particular extensions, e.g. if you don't want the user to be able to paste images (or colored text, or tables, ...).

I was initially in favour of settings minimal defaults (e.g. starterkit), but after having to explain a colleague that he had to add color first as an extension, and next as a control. And if he forgot to add the extension, the control wouldn't work, I started to learn towards your idea of adding them all by default.

AnnMarieW:
Alright - docs are updated.  Thanks for fixing the test too!  

@alexcjohnson any final comments?  It think this is good to go.

alexcjohnson (approved):
💃

AnnMarieW:
This is so exciting!  I'll work on getting 1.1.0 released tomorrow.  

@emilhe and @alexcjohnson - Thanks so much--it was fun working on this new feature with both of you :grin:

Issue #534, opened 2025-03-15
[Feature Request] Add wildcard search to MultiSelect


omarirfa:
## Description
Currently, the MultiSelect component only supports exact substring matching in search, which limits its flexibility when working with large datasets or specific patterns. For my usecase, I sometimes have data that I would like to select all 10* or abc* and current multiselect does not seem to handle that. I would have to use the textinput to do wildcard searches with the backend code handling the search. 

## Proposed solution
Add a new `wildcard` boolean prop that enables wildcard search using asterisks (*) as wildcard characters.

## Use cases
- Searching for items with specific patterns (e.g., "10*" to find all items starting with 10)
- Filtering alphanumeric codes (e.g., ".060x*" to find all items starting with .060x)
- Enabling more flexible searches for end users without complex regex knowledge

## Proposed solution
Heres some suggested code on how the implementation can look like, I based it on the MultiSelect.tsx:

```
import { MultiSelect as MantineMultiSelect } from "@mantine/core";
import { useDebouncedValue, useDidUpdate, useFocusWithin } from "@mantine/hooks";
import { BoxProps } from "props/box";
import { ComboboxLikeProps } from "props/combobox";
import { DashBaseProps, PersistenceProps, DebounceProps } from "props/dash";
import { __ClearButtonProps } from "props/button";
import { __BaseInputProps } from "props/input";
import { ScrollAreaProps } from "props/scrollarea";
import { StylesApiProps } from "props/styles";
import React, { useState } from "react";
import { filterSelected } from "../../../utils/combobox";
import { setPersistence, getLoadingState } from "../../../utils/dash3";

interface Props
    extends BoxProps,
        __BaseInputProps,
        ComboboxLikeProps,
        StylesApiProps,
        DashBaseProps,
        DebounceProps,
        PersistenceProps {
    /** Controlled component value */
    value?: string[];
    /** Controlled search value */
    searchValue?: string;
    /** Maximum number of values, `Infinity` by default */
    maxValues?: number;
    /** Determines whether the select should be searchable, `false` by default */
    searchable?: boolean;
    /** Message displayed when no option matched current search query, only applicable when `searchable` prop is set */
    nothingFoundMessage?: React.ReactNode;
    /** Determines whether check icon should be displayed near the selected option label, `true` by default */
    withCheckIcon?: boolean;
    /** Position of the check icon relative to the option label, `'left'` by default */
    checkIconPosition?: "left" | "right";
[...]

PR #535, opened 2025-03-16
Multiselect filter integration

omarirfa:
## Description
This is an attempt to solve issue #534 with integrating filter and allowing the functionality to be the same as mantine upstream while also, making sure this complies with the dash mantine roadmap.

I have not tested these completely,  heres how the user would be able to run this:

## Changes

1. Added stringToFunction utility to convert string filter code to executable functions
2. Added filter prop to ComboboxProps interface
3. Updated MultiSelect to support custom filter functions via comboboxProps.filter
4. Filters are executed within a try/catch block for safety

## Solution usage
Heres how users will use this functionality:
```
import dash
from dash import html, _dash_renderer


app = dash.Dash(__name__)

wildcard_filter = """
const itemValue = typeof item === 'string' ? item : item.value || item.label || '';
const itemLabel = (typeof item === 'object' && item.label) ? item.label : itemValue.toString();

if (search.includes('*')) {
    const pattern = search.replace(/\\*/g, '.*');
    const regex = new RegExp(pattern, 'i');
    return regex.test(itemLabel);
}

return itemLabel.toLowerCase().includes(search.toLowerCase());
"""

    dmc.MultiSelect(
        id="test-select",
        data=["Option 10", "Option 100", "Option 1000"],
        label="Test Wildcard Filter",
        placeholder="Try '10*' or '*Thing'", 
        searchable=True,
        comboboxProps={"filter": wildcard_filter}
    ),
    html.Div(id="output")
])

if __name__ == '__main__':
    app.run_server(debug=True)
```

## Potential Test Cases

- [x] Base functionality works without filter
- [x] Custom filter function can be passed and executed
- [x] Wildcard filter works with patterns (10*, thing, az)
- [x] Filter handles edge cases (empty search, malformed patterns) 

I would love your thoughts on this @AnnMarieW if this is a more adequate solution!

edit: my current use case seems to be i want all the values found after the wildcard/filter and then the user should have a button to select them all or some way to grab all those relevant values. will need to see what is the equivalent to that on mantine

edit 2 : it seems a functionality like that does not exist in mantine, it was discussed a bit in the github discussions https://github.com/orgs/mantinedev/discussions/3635 on mantinedev. Any thoughts on if we can add the select all feature here to multiselect?

### DIFF
--- a/src/ts/components/core/combobox/MultiSelect.tsx
+++ b/src/ts/components/core/combobox/MultiSelect.tsx
@@ -7,8 +7,8 @@ import { __ClearButtonProps } from "props/button";
 import { __BaseInputProps } from "props/input";
 import { ScrollAreaProps } from "props/scrollarea";
 import { StylesApiProps } from "props/styles";
-import React, { useState } from "react";
-import { filterSelected } from "../../../utils/combobox";
+import React, { useState, useEffect } from "react";
+import { filterSelected, stringToFunction } from "../../../utils/combobox";
 import { setPersistence, getLoadingState } from "../../../utils/dash3";
 
 interface Props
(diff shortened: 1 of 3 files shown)


AnnMarieW:
Hi @omarirfa 

Thanks for this PR. This is a great start!

Adding functions as props is part of a much larger project. The API design needs to work consistently across all props that accept functions in the Mantine codebase. There are also important security considerations around executing functions safely—one of the key reasons Dash does not support this by default.

I’m still working through the best approach for this feature, so it may take some time before we can move forward. I really appreciate your patience in the meantime.

omarirfa:
No worries. Makes sense regarding the security concerns. I look forward to what would be the best approach

AnnMarieW:
@omarirfa 
I just got started with a PR that enables functions as props.  I'd love to get your feedback on https://github.com/snehilvj/dash-mantine-components/pull/580


There is an example of a custom filter with a `Select` component - and it also works with the `MultiSelect`

AnnMarieW:
@omarirfa I'm going to close this in favor of #580.   Thanks for this PR and the comments on 580.  I look forward to working with you on this feature over there.

PR #538, opened 2025-03-18
Fix unselect not working in RadioGroup

emilhe:
Fix test failure. Seemed like a race condition in RadioGroup. By changing the Dash update invocation to only happen on value changes, the test passes every time. Behaviour validated manually also.


AnnMarieW:
Thanks for this @emilhe 
This was included with #530, so I'll close this now.

Issue #525, opened 2025-03-04
Rework Notifications into `NotificationContainer` and Expose notifications API from Mantine


BSd3v:
We need to rework the `NotificationProvider` and `Notifications` components into another `NotificationContainer` and expose the underlying `notifications` API from `Mantine`. This change will help us better manage notifications and provide more flexibility in how they are displayed and handled.

Tasks:
- [ ] Rework the `NotificationProvider` component into `NotificationContainer`. (`notifications` stream into the container and work with the queue to ensure all notifications are from callbacks are displayed before being removed)
    - [ ] expose props for api (`clean`, `cleanQueue`)
    - [ ] `notifications` handle for `{add, update, hide}`?
- [ ] Expose the underlying notifications API from Mantine. (`dash_mantine_components.notification_container` similar to `dash_ag_grid.getApi`)
- [ ] Update documentation to reflect these changes. (deprecate `NotificationProvider` and `Notifications` for `v1` and remove in `v2`)

Additional Context:
The current `Notifications` component has limitations in terms of flexibility and customization. This is also not a good pattern for the notifications as Mantine itself only has the single component. The components are also stuck in the dash eco-system unless replaced or removed. (eg. `Patch` never removes) By exposing the notifications API from Mantine, we can leverage its full capabilities and provide a more robust solution for managing notifications.

AnnMarieW:
Thanks for the write-up @BSd3v 
For a little more background see #523 and #522

PR #539, opened 2025-03-20
adding `NotificationContainer` to be more inline with `Mantine`

BSd3v:
utilizes the api more closely (https://mantine.dev/x/notifications/)
exposes the api and store at `appNotifications` in the `dash_mantine_components` namespace

fixes https://github.com/snehilvj/dash-mantine-components/issues/525


BSd3v:
I cant figure out how to tie the store together to be able to see what's in the queue vs the active notifications.

https://mantine.dev/x/notifications/#subscribe-to-notifications-state

BSd3v:
Here is an example app to show the difference in the API:
```

from dash import Output, Input, html, callback, Dash, ctx, _dash_renderer
from dash.exceptions import PreventUpdate

# this is optional when using dash >3.0.0

app = Dash(__name__, external_stylesheets=dmc.styles.ALL)

    [
        dmc.NotificationContainer(id='test-container'),
        dmc.Group(
            children=[
                dmc.Button(
                    "Load Data",
                    id="show-notification",
                ),
                dmc.Button(
                    "Update",
                    id="update-notification",
                ),
            ],
        ),
    ]
)


@callback(
    Output("test-container", "sendNotifications"),
    Input("show-notification", "n_clicks"),
    Input("update-notification", "n_clicks"),
    prevent_initial_call=True,
)
def notify(nc1, nc2):
    if not ctx.triggered:
        raise PreventUpdate
    else:
        button_id = ctx.triggered_id
        if "show" in button_id:
            return {'show': [
                dict(
                    id=f"my-notification-{nc1}",
                    title="Process initiated",
                    message="The process has started.",
                    loading=True,
                    color="orange",
                    autoClose=False,
                )
                ]
            }
        else:
            return {'update': [
                dict(
                    id=f"my-notification-{nc2}",
                    title="Data loaded",
                    message="Notification closing in 2 seconds",
                    color="green",
                    loading=False,
                    action="update",
                    autoClose=2000,
                    icon=DashIconify(icon="akar-icons:circle-check"),
                )
                ]
            }


if __name__ == "__main__":
    app.run(debug=True)
```

BSd3v:
Addresses https://github.com/snehilvj/dash-mantine-components/issues/525

BSd3v:
fixes https://github.com/snehilvj/dash-mantine-components/issues/525

AnnMarieW:
Hey Alex - this PR is ready for review.  It works well and looks like it resolves the issue from the original component.

Would like to get your thoughts on the API and how to handle the transition from the current method to this one in order to avoid breaking changes (and confusion)

src/ts/components/extensions/notifications/NotificationContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+        }
+    }, [clean, cleanQueue]);
+
+    useEffect(() => {
+        appNotifications['api'] = notifications
+        appNotifications['store'] = notificationsStore.getState
+        return () => {
+            delete appNotifications['api']
+            delete appNotifications['store']
+        }
```
alexcjohnson, line 128: Is it important to remove these on unmount? Thinking about two scenarios:
1. A clientside callback that's set to perhaps clear the notifications in several situations, one of which also removes the `NotificationContainer` entirely, depending on the order of operations the clientside callback may be executed after the `NotificationContainer` has already unmounted. Feels like we could leave the `api` and `store` there even if accessing them isn't going to do anything visible.
2. Changing layouts/pages, if the `NotificationContainer` is in both layout contents rather than in the page skeleton. Can we always guarantee that the new one will be created after the previous one's unmount hooks have executed? If for any reason these happened in the opposite order we'd end up deleting them after creating them the second time and they'd end up gone. This one seems unlikely, unless you do something really weird like put the two layout chunks into different parts of the page skeleton so they're both briefly on screen simultaneously... nevertheless I feel like we could just leave these items where they are and avoid this possibility.

src/ts/components/extensions/notifications/NotificationContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+        if (clean || cleanQueue) {
+            if (clean) {
+                notifications.clean()
+            }
+            if (cleanQueue) {
+                notifications.cleanQueue()
+            }
+            setProps({clean: false, cleanQueue: false})
+        }
+    }, [clean, cleanQueue]);
```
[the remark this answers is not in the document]
BSd3v: We could, to make the it one function to one api call. 

I had them together so that you could stack the `setProps` to keep the rerenders down.

src/ts/components/extensions/notifications/NotificationContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+    /** Notification `max-height`, used for transitions, `200` by default */
+    notificationMaxHeight?: number | string;
+    /** Maximum number of notifications displayed at a time, other new notifications will be added to queue, `5` by default */
+    limit?: number;
+    /** Notifications container z-index, `400` by default */
+    zIndex?: string | number;
+    /** Determines whether notifications container should be rendered inside `Portal`, `true` by default */
+    withinPortal?: boolean;
+    /** Notifications to be passed to the API */
+    sendNotifications?: NotificationsFormat;
```
alexcjohnson, line 81: Overall I like this pattern, but most of the time you're just going to use `{'show': [...]}` - I wonder if it would be better as three separate props, like `showNotifications, hideNotifications, updateNotifications`?

src/ts/components/extensions/notifications/NotificationContainer.tsx
```diff
@@ -0,0 +1,140 @@
...
+    /** Notifications API: removes all notifications from the notifications state and queue*/
+    clean?: boolean;
+    /** Notifications API: removes all notifications from the queue*/
+    cleanQueue?: boolean;
+}
+
+/** NotificationContainer */
+const NotificationContainer = (props: Props) => {
+    const { setProps, loading_state, sendNotifications, clean, cleanQueue, ...others } = props;
+    const [appNotificationStore, setAppNotificationStore] = useState(notificationsStore)
```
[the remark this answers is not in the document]
BSd3v: The notification store?

I thought about using something like this: https://mantine.dev/x/notifications/#subscribe-to-notifications-state

I think I opted for exposing the api instead.

src/ts/components/extensions/notifications/NotificationContainer.tsx
BSd3v: Yeah, we could do that.

AnnMarieW:
Also note that Mantine is releasing V8 next month.  If the best solution has breaking changes, we could include them with the others when we upgrade to V8.

src/ts/components/extensions/notifications/NotificationContainer.tsx
BSd3v: So, when working with this, its kinda frustrating have to split this into multiple props. I'll push the changes and let you look at it.

BSd3v:
The notifications wont be breaking changes as this is a new component. We should really deprecate the old version though.

src/ts/components/extensions/notifications/NotificationContainer.tsx
alexcjohnson: Not recommended is different from won't happen 😉

The user's goal in these scenarios is not to have multiple notification providers at the same time, it's to move the notification provider from one place to another - perhaps on one page you want them at the top level, so extending vertically down from the top right corner on top of other content, but on a different page you want something else always visible in the top right corner so you put the notifications inside a smaller container.

BSd3v: The notification itself can be placed in different places, or you could have a clientside update based upon the layout you are loading. :)

src/ts/components/extensions/notifications/NotificationContainer.tsx
alexcjohnson: Frustrating when you have logic that needs to decide between show and update or some such? I can see that, though I still feel like this is an edge case so I'm not too bothered by that being less convenient. Do you agree or do you think it's a common need?

An alternative would be to make one list of notifications, but give each one an extra optional key like `action ?= 'show', 'update', 'hide'`. That means the object no longer quite matches the Mantine notification format, but one extra key isn't so bad. Then the code to handle them would be something like:
```js
useEffect(() => {
    if (notifications) {
        notifications.forEach((notification) => {
            const {action, ...realNotification} = notification;
            notifications[action || 'show'](newRenderDashComponents(realNotification, ['message', 'icon', 'title']));
        });
        setProps({ notifications: [] }); // Avoid duplicate processing
    }
}, [notifications]);
```

BSd3v: That is how the original dmc api was done. 

I think it feels better this way, especially with the default of show.

AnnMarieW: Actaully, I also think it would be better to have separate props for  `show` `update` and `hide` (rather than `sendNotificaitons`) It's a closer match to the Mantine API.  It also makes it easier to know what props each action requires.   For example, `hide` only needs the notification `id`,  `show` and `update` can take any of the valid `Notification` props.  Currently if you use an invalid `action` you get a very hard to debug error message.  Separate props would ensure only the valid Mantine function is used.  

Bryan has made a good case that it's more convenient if it's all in one prop (`sendNotifications`), but a workaround for that could be using `setProps` or the api directly in a clientside callback using the new `appNotifications`

src/ts/components/extensions/notifications/NotificationContainer.tsx
```diff
@@ -0,0 +1,126 @@
...
+    /** Notification width, cannot exceed 100%, `440` by default */
+    containerWidth?: number | string;
+    /** Notification `max-height`, used for transitions, `200` by default */
+    notificationMaxHeight?: number | string;
+    /** Maximum number of notifications displayed at a time, other new notifications will be added to queue, `5` by default */
+    limit?: number;
+    /** Notifications container z-index, `400` by default */
+    zIndex?: string | number;
+    /** Determines whether notifications container should be rendered inside `Portal`, `true` by default */
+    withinPortal?: boolean;
```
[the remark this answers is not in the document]
BSd3v: `store` as in what? Is this to be from the `notifications`? If so, we'd have to have a hook from the store to know when it updated.

src/ts/components/extensions/notifications/NotificationContainer.tsx
```diff
@@ -0,0 +1,126 @@
...
+    cleanQueue?: boolean;
+}
+
+/** NotificationContainer */
+const NotificationContainer = (props: Props) => {
+    const { setProps, loading_state, sendNotifications, clean, cleanQueue, ...others } = props;
+
+    const componentPath = getContextPath()
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
BSd3v: `useDidUpdate` only runs on the prop updates, this would not capture `sendNotifications` initially sent to the component.

src/ts/components/extensions/notifications/NotificationContainer.tsx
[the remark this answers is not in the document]
AnnMarieW: Ok, let's skip the `store` prop

src/ts/components/extensions/notifications/Notification.tsx
```diff
@@ -42,6 +42,7 @@ const Notification = (props: Props) => {
     const { action, setProps, loading_state, ...others } = props;
 
     useEffect(() => {
+        console.warn('this method of Notifications is deprecated and will be removed in the next major release')
```
AnnMarieW, line 45:
```suggestion
        console.warn('This method of Notifications is deprecated and will be removed in a future major release.  Instead, use `NotificationContainer')
```

src/ts/components/extensions/notifications/NotificationProvider.tsx
```diff
@@ -34,6 +34,10 @@ interface Props extends BoxProps, StylesApiProps, DashBaseProps {
 const NotificationProvider = (props: Props) => {
     const { setProps, loading_state, ...others } = props;
 
+    useEffect(() => {
+        console.warn('this method of Notifications is deprecated and will be removed in the next major release')
```
AnnMarieW, line 38:
```suggestion
        console.warn('This method of Notifications is deprecated and will be removed in a future major release.  Instead, use `NotificationContainer')
```

src/ts/utils/dash3.ts
```diff
@@ -16,6 +16,23 @@ export const isDash3 = (): boolean => {
...
+    }
+
+    const stringifyVal = (v: any) => (v && v.wild) || JSON.stringify(v);
+
+    const parts = Object.keys(id)
+        .sort()
+        .map((key) => JSON.stringify(key) + ':' + stringifyVal(id[key]));
+
+    return '{' + parts.join(',') + '}';
+};
```
BSd3v, line 33: This should check if `dash3`, if `dash3` it should use `dash_component_api.stringyId` instead.

BSd3v:
Should add a couple tests for `hideNotifications` and also checking with a couple of pattern-matching callbacks.

AnnMarieW:
Here are the new Notification docs https://github.com/snehilvj/dmc-docs/pull/199

AnnMarieW:
@alexcjohnson This PR is ready for a final review.  We are planning on including this in V2 so that people can refer to the V1 docs if they are still using the deprecated components.

The new Notification docs and a detailed migration guide are ready  (with some final edits to do)

The only open item is whether to rename the component `NotificationsContainer`  (with an s)  any thoughts on that?

alexcjohnson (approved):
LGTM! I'd stick with `NotificationContainer`, I see the point about it holding multiple notifications, but it's grammatically funny. A shoe store sells many shoes, a bike lane can have many bikes...

src/ts/components/extensions/notifications/NotificationProvider.tsx
```diff
@@ -34,6 +34,10 @@ interface Props extends BoxProps, StylesApiProps, DashBaseProps {
 const NotificationProvider = (props: Props) => {
     const { setProps, loading_state, ...others } = props;
 
+    useEffect(() => {
+        console.warn('This method of Notifications is deprecated and will be removed in a future major release.  Instead, use `NotificationContainer')
```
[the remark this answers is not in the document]
AnnMarieW: Good point - that's better.  But should probably be `NotificationProvider` and `Notification` are deprecated....

Issue #542, opened 2025-03-21
[bug] Issue with dmc.Notification when switching pages on DMC 1.0+


benweinberg89:
Another issue I am noticing upon upgrading my DMC from 0.15.3 to 1.0 or 1.1, is an issue with dmc Notifications in my multipage DMC app. On my first page, there are callbacks that yield a dmc.Notification to a div. Here is an example of one notification:

within AppShellMain():

```py
html.Div(
    id="notifications-container",
)
```

callback:

```py
Output("notifications-container", "children", allow_duplicate=True),
...
return dmc.Notification(
            title="Success",
            message=f"Opened experiment: {title}",
            color="green",
            action="show",
            autoClose=2000,
            icon=DashIconify(icon="ic:round-check-circle"),
        )
```

app.py where i have the notification provider setup:
```py
dmc.MantineProvider(
        children=[
            dmc.NotificationProvider(position="bottom-right"),
            html.Div(id="notifications-container"),
            dmc.AppShell(
                id="app-shell",
                children=[
                    app_header,
                    app_navbar,
                    html.Div(id="app-content", children=[dash.page_container]),
                ],
```

I'm noticing in DMC1.0/1.1 that when I go to other pages before that Notification autocloses, I get an error that crashes the page once the Notification autocloses. It doesn't really matter which page I go to (except if I stay on the original page and wait for it to close and then there is no issue), I get a similar error and it lists out a large Json-like structure (attached below is for one page) that maps to whatever is on that page. Running my same exact code on DMC 0.15.3 never gave me this issue. Importantly, when I disable the Notifications from the app, I don't have this issue anymore. I am using Dash 2.18.2.

```
(This error originated from the built-in JavaScript code that runs Dash apps. Click to see the full stack trace or open your browser's console.)
Error: An object was provided as `children` instead of a component, string, or number (or list of those). Check the children property that looks something like:

```

```json
{
"props": {
"children": [
{
"props": {
"children": [
{
"props": {
"children": {
"props": {
"children": [
null,
{
"props": {
"children": {
"props": {
"action": "hide"
}
}
}
}
]
}
}
}
}
]
}
}
]
}
}
```

```
at validateComponent (http://127.0.0.1:8050/_dash-component-suites/dash/dash-renderer/build/dash_renderer.v2_18_2m1733601907.dev.js:3342:11)
[...]

PR #546, opened 2025-03-24
adjustments to notification to test dash store if it exists in the dash tree

BSd3v:
fixes https://github.com/snehilvj/dash-mantine-components/issues/542

@alexcjohnson what do you think about this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Change Log
 
+# unreleased
+
 # 1.1.0
 
 ### Added
--- a/src/ts/components/extensions/notifications/Notification.tsx
+++ b/src/ts/components/extensions/notifications/Notification.tsx
@@ -4,6 +4,7 @@ import { BoxProps } from "props/box";
 import { DashBaseProps } from "props/dash";
 import { StylesApiProps } from "props/styles";
 import React, { useEffect } from "react";
+import { getChildLayout, getContextPath, isComponentInDash } from "../../../utils/dash3";
 
 interface Props extends BoxProps, StylesApiProps, Omit<DashBaseProps, "id"> {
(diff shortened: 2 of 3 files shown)


AnnMarieW:
Here's a sample app.  Note that the notification container is only in the page1 layout.  Before this change if you switched pages while the notification was still visible, it would cause an error.





```python
import dash
from dash import Dash, _dash_renderer, html, callback, Input, Output

app = Dash(external_stylesheets=dmc.styles.ALL, use_pages=True, pages_folder="")


page1_layout = html.Div(
    [
        dmc.NotificationProvider(),
        html.Div(id="notifications-container"),
        dmc.Button("Show Notification", id="notify"),
    ]
)

dash.register_page("home", path="/", layout=html.Div("I'm home"))
dash.register_page("page1", path="/page-1", layout=page1_layout)


component = dmc.Box([
    dmc.NavLink(label="home", href="/", active='exact'),
    dmc.NavLink(
            label="Page 1",
            childrenOffset=28,
            href="/page-1",

    ),
    dmc.Divider(mb="lg"),
    dash.page_container
])


@callback(
    Output("notifications-container", "children"),
    Input("notify", "n_clicks"),
    prevent_initial_call=True,
)
def show(n_clicks):
    return dmc.Notification(
        title="Hey there!",
        id="simple-notify",
        action="show",
        message="Notifications in Dash, Awesome!",
        icon=DashIconify(icon="ic:round-celebration"),
    )

if __name__ == "__main__":
    app.run(debug=True)
```

AnnMarieW:
Looks like this doesn't fix the issue.  Try running with this example https://github.com/snehilvj/dash-mantine-components/issues/542#issuecomment-2755733212

AnnMarieW:
@alexcjohnson Just FYI - this doesn't seem to solve the issue either. We think it's best to close this issue and revert #523.  This will fix the issues reported in #542.

The  PR #539 is likely the better solution so we can continue development over there.

PR #567, opened 2025-04-20
bundle optional CSS

### LINKED ISSUE #552: [Feature Request] Bundle optional CSS files
### Feature Request: Bundle Optional Stylesheets with DMC  

DMC currently requires on additional  stylesheets for certain component.  The recommended way to add these is to use external stylesheets  which can cause issues if the CDN is down or the app needs to run offline.  While it's a best practice to also include the stylesheets in the assets folder, it would be more convenient to bundle them with DMC:  

1. Apps remain functional without internet access.  
2.  Avoids breakage if the CDN is unavailable.  
3. Ensures styles match the DMC version without manual updates.  

Total bundle size increase is about 44k.  However, currently RichTextEditor, Charts and CodeHighlight components are lazy loaded, so including the CSS in the components won't have an impact on initial load time if they are not used on the page.

Image

AnnMarieW:
closes #552 

Adds 20KiB to  `dash_mantine_components.js`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+# Unreleased
+
+### Added
+
+- Optional component stylesheets are now bundled automatically. It’s no longer necessary to include them as external stylesheets or add them to the `assets` folder. PR #567 by @AnnMarieW
+
+
 # 1.1.2rc1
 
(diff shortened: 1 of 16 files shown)


alexcjohnson (approved):
💃 Makes sense to me!

PR #568, opened 2025-04-20
add inputProps

### LINKED ISSUE #547: Feature Request - Add maxLength to dmc.TextInput
For parity with dcc.Input:

> **maxLength** (string | number; optional): If the value of the type attribute is text, email, search, password, tel, or url, this attribute specifies the maximum number of characters (in UTF-16 code units) that the user can enter. For other control types, it is ignored. It can exceed the value of the size attribute. If it is not specified, the user can enter an unlimited number of characters. Specifying a negative number results in the default behavior (i.e. the user can enter an unlimited number of characters). The constraint is evaluated only when the value of the attribute has been changed.

Sample app with current behavior and two workarounds:
1. Using a callback to limit the length of the value - for better performance, use clientside callback with set_props
2. Using dmc.InputWrapper - the styling won't be exactly the same, but it allows the inclusion of other dmc properties like label or description. 

https://github.com/user-attachments/assets/a7bc0636-a884-4fd8-a13a-56f0ce42985c

> dash                          2.18.2
dash_mantine_components       1.0.0

```
from dash import Dash, dcc, html, Input, Output, callback, MATCH
import dash

# add this before creating app object

app = Dash(external_stylesheets=[dmc.styles.ALL])

app.layout = html.Div(dmc.MantineProvider(
    [
        html.I("dcc.Input with maxLength=3"),
        html.Br(),
        dcc.Input(id={"type":"input", "index":"dcc"}, type="text", maxLength=3, style={'marginRight':'10px'}),
        html.Div(id={"type":"output", "index":"dcc"}, children="No text yet"),
        #####
        dmc.TextInput(id={"type":"input", "index":"dmc"}, label="dmc.TextInput", description="Some description"),
        html.Div(id={"type":"output", "index":"dmc"}, children="No text yet"),
        ####
        dmc.TextInput(id={"type":"input", "index":"dmc+workaround"}, label="dmc.TextInput with callback workaround", description="Some description"),
        html.Div(id={"type":"output", "index":"dmc+workaround"}, children="No text yet"),
        ####
        dmc.InputWrapper(
            dcc.Input(
                id={"type":"input", "index":"dcc+dmc"}, type="text", maxLength=3,
                ),
            label="dmc.InputWrapper", description="dcc.Input with maxLength=3 wrapped in dmc.InputWrapper",
        ),
        html.Div(id={"type":"output", "index":"dcc+dmc"}, children="No text yet"),
    ]
), style={"margin":"20px"})

@callback(
    Output({"type":"output", "index":MATCH}, "children"),
    Input({"type":"input", "index":MATCH}, "value"),
    prevent_initial_call=True
)
def update_output(text):
    return f"Text is: {text}"

@callback(
    Output({"type":"input", "index":"dmc+workaround"}, "value"),
    Input({"type":"input", "index":"dmc+workaround"}, "value"),
    prevent_initial_call=True
)
def set_limit(value):
    LIMIT=3
    if len(value) > LIMIT:
        return value[:LIMIT]

if __name__ == "__main__":
    app.run(debug=True)

```

AnnMarieW:
closes #547 

This PR addes `inputProps` to make it possible to pass any valid attribute to the `input` element.

For example as requested in issue 547, you can include maxlength like this:

```python
dmc.TextArea(inputProps={"maxLength": 3})

```

### DIFF
--- a/src/ts/components/core/input/JsonInput.tsx
+++ b/src/ts/components/core/input/JsonInput.tsx
@@ -64,13 +64,13 @@ const JsonInput = ({
 
     return (
         <MantineJsonInput
+            {...inputProps}
             data-dash-is-loading={getLoadingState(loading_state) || undefined}
             onChange={setVal}
             value={val}
             onKeyDown={handleKeyDown}
             onBlur={handleBlur}
             autoComplete={autoComplete}
-            {...inputProps}
             {...others}
         />
     );
--- a/src/ts/components/core/input/JsonInput.tsx
+++ b/src/ts/components/core/input/JsonInput.tsx
@@ -28,6 +28,7 @@ const JsonInput = ({
     n_blur = 0,
     debounce = false,
     autoComplete = "off",
+    inputProps,
     ...others
 }: Props) => {
 
@@ -69,6 +70,7 @@ const JsonInput = ({
             onKeyDown={handleKeyDown}
             onBlur={handleBlur}
             autoComplete={autoComplete}
+            {...inputProps}
             {...others}
         />
     );
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Change Log
 
+# Unreleased
+
+### Added
+
+- Added `inputProps` to JsonInput, NumberInput, PasswordInput, TextInput, and TextArea components to allow passing props directly to the underlying input element. #568 by @AnnMarieW
+
+
 # 1.1.2rc1
 
 ### Fixed
--- a/src/ts/components/core/input/NumberInput.tsx
+++ b/src/ts/components/core/input/NumberInput.tsx
@@ -114,13 +114,13 @@ const NumberInput = ({
 
     return (
         <MantineNumberInput
+            {...inputProps}
             data-dash-is-loading={getLoadingState(loading_state) || undefined}
             onChange={setVal}
             value={val}
             onKeyDown={handleKeyDown}
             onBlur={handleBlur}
             autoComplete={autoComplete}
-            {...inputProps}
             {...others}
         />
     );
--- a/src/ts/components/core/input/NumberInput.tsx
+++ b/src/ts/components/core/input/NumberInput.tsx
@@ -77,6 +77,7 @@ const NumberInput = ({
     n_blur = 0,
     debounce = false,
     autoComplete = "off",
+    inputProps,
     ...others
 }: Props) => {
 
@@ -119,6 +120,7 @@ const NumberInput = ({
             onKeyDown={handleKeyDown}
             onBlur={handleBlur}
             autoComplete={autoComplete}
+            {...inputProps}
             {...others}
         />
     );
--- a/src/ts/components/core/input/PasswordInput.tsx
+++ b/src/ts/components/core/input/PasswordInput.tsx
@@ -36,6 +36,7 @@ const PasswordInput = ({
     n_blur = 0,
     debounce = false,
     autoComplete = "off",
+    inputProps,
     ...others
 }: Props) => {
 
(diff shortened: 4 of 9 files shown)


AnnMarieW:
Here is a sample app to try:  [pycafe](https://py.cafe/snippet/dash/v1?#c=H4sIAKJZBWgEA8VVXYvbOhD9K0b3ZS_ESmwnziZguJ9wH3qhlEIf1otRbCXWRpZUSd6sd8l_75GdTSm0bN-avGQ058ycGc0oL6TWDSdbIjqjrY8a5tqqY8oLxata41Bx5V3EXNR0dan2VncjKLoQ_sHvUoUvMyYqRvuGP3luFZOV84PkruXcuwJ8Otn0z3fvfg-UawIQg_sjiMxydlOqCB-hTO_fW21c8VKSjj1Jrg6-Lck2ys6zCSPZjsuiJP8qpIw8ApTk4mq4q60wXmgFAOgR-L6N9D7KorplltVeWzfiRzljDVSyQfeviv6fWgERj6Lh9ubuKvn-QhH7qKoU63hVRUURlaRC_4SqKsicdISgtlc3Dd_1h-Kj7TmoZEYs_9wLyzsEc7iB0NVS_egC_oha743bzudmoDXb8_mhnTPrxR5VzJ3irZCPD_NAjy_0-Ct9nm7W-SrdZMt0hHwnQ5zQhKa2TmIzZLHCYczUQE-thFQ_mDAkgQrLDL7VCrYZdIO2xI8LmuY0gWtqHtm-kEduHTpPtikq1dp_0KdwfIlkYc1IDdGN5QDdXT2e7Rz3cJ5E41uyTVaLGemE-jSZ2WT9x8WhRZ5giga0vZD8L0R13P6tlccNcPuDDAEa7yYsIIaFuISc78-zK-aq4nsB4PzKC7drhm_Z9u3q3q7oovCtal4L-UbANU17iZqsVz-TEzuEYybfSvqKC1nD9zwbu4oxvrvHEOj6GMwQBaMF_G9Ri9HB2q5Xt3m2WPJ8t0n3q5ov85wtd3uW3Obr9Sor1U4KdeS2KBK6oQs8ETzMuCiKdJEuaUpTHGF1UR1m1HZMiueAzmg2uqSoj0Vxi1FeT7sUXAuKwL9sffBoSuag6iJENIoFAwKnN1SKXdxxzxrm4bil2CS4vGuYOnCre4fiUTma8SDUA0sDOQmROmaPvXGoJiASuiqVYvbUMgkKEKvAUdz5mLlB1UKH0zwcGqm9HIoCRkj2usehR0WBZQ6H4XkCd8yeJYFlubeDUIcxOF2WCrfQG691yJdv6CrQnHgK_mTMg3EBPg7_B3BoBdySJmOw3koUnk21gXfi9vjM8UCOfcpL9SwMxOBaR7hzaCQ5fwERpCQ2sAYAAA)

src/ts/components/core/input/JsonInput.tsx
```diff
@@ -69,6 +70,7 @@ const JsonInput = ({
             onKeyDown={handleKeyDown}
             onBlur={handleBlur}
             autoComplete={autoComplete}
+            {...inputProps}
             {...others}
```
alexcjohnson, line 74: Your goal here is to pass arbitrary props all the way through to the HTML, which makes sense. I'm just thinking about what happens if the user puts things in `inputProps` that are already supplied elsewhere. As written, if you have the same keys in both `inputProps` and `others`, `others` wins because it comes later, and I think that makes sense. But `inputProps` can also override the explicitly named props before it, like `value` or `onChange`, which would be rather awkward! We could of course say it's obvious you shouldn't do this, but if it's as easy as moving `{...inputProps}` up to the top of the prop list to prevent this pitfall maybe we should?

AnnMarieW: Yes, good point.  I'll make the change

alexcjohnson (approved):
💃

PR #580, opened 2025-05-02
allow functions as props

### LINKED ISSUE #356: [Feature Request] Allow functions as props.
First of - thanks for this awesome library! It's a lot of fun to work with and it really improves the look of the Dash Apps it is used with 😃.

I have a feature request to add support for label formatting on slider elements, e.g., `40 °C` for values of `40`.

### Background
This is supported in `Mantine` (see https://mantine.dev/core/slider/#control-label) and would be great to have in `dmc`. The [docs](https://www.dash-mantine-components.com/components/slider#slider) hint at this being implemented already with the `label` argument (_Function to generate label or any react node to render instead, set to None to disable label._), but I couldn't get it to work or find an example that explains this.

### What I've tried
I guess from Python the most straight-forward way to implement this would be to support passing functions to the `label` argument of the `Slider` constructor. Currently, passing `label=lambda x: x` raises an `Error loading layout` error.

Thanks!

AnnMarieW:
closes #356 

This is the start of allowing function to be passed from a dash app.

In this example the `Slider`  `label` prop can be a function:
> label:  ReactNode | ((value: number) => ReactNode)

It's possible to pass a function that is defined in a `.js` file in `assets` like this:
 ` label={"function": "myLabel"}`

```python

from dash import Dash

app = Dash()

    dmc.Slider(value=25, p=100, label={"function": "myLabel"})
)

if __name__ == "__main__":
    app.run(debug=True)

```

.js file in /assets:
```js
var dmcfuncs = window.dashMantineComponentsFunctions = window.dashMantineComponentsFunctions || {};

dmcfuncs.myLabel = function(value) {
            return `${value} °C`
        }

```

image

### DIFF
--- a/src/ts/utils/prop-functions.ts
+++ b/src/ts/utils/prop-functions.ts
@@ -1,0 +1,78 @@
+/**
+ * Utility to allow passing a function as a prop from a dash app.
+ * For example label={'function': 'myLabel'}
+ * where 'myLabel' is a function defined in .js file in /assets
+ *
+ */
+
+export const context = {
+//             d3,
+//             dash_clientside,
+//             ...customFunctions,
+           ...(window as any).dashMantineComponentsFunctions
+        };
+
+
+function isPlainObject(o) {
+   return (o === null || Array.isArray(o) || typeof o == 'function' || o.constructor === Date ) ?
+           false
+          :(typeof o == 'object');
+}
+
+function isFunction(functionToCheck) {
+   return functionToCheck && {}.toString.call(functionToCheck) === '[object Function]';
+}
+
+
+export function resolveProp(prop, context = {}) {
+    // If it's not an object, just return.
+    if (!isPlainObject(prop)){
+        return prop
+    }
+    // Check if the prop should be resolved a variable.
+    if (prop.function){
+        return resolveVariable(prop, context)
+    }
+    // Check if the prop should be resolved as an arrow function.
+    if (prop.arrow){
+        return (...args) => prop.arrow
+    }
+    // If none of the special properties are present, do nothing.
+    return prop
+}
+
+function resolveVariable(prop, context){
+    const variable = getDescendantProp(window, prop.function)
+
+    // If it's not there, raise an error.
+    if(variable === undefined){
+        throw new Error("No match for [" + prop.function + "] in the global window object.")
+    }
+    // If it's a function, add context.
+    if(isFunction(variable) && context){
+        return (...args) => variable(...args, context)
+    }
+    // Otherwise, use the variable as-is.
+    return variable
+}
+
+function getDescendantProp(obj, desc, defaultPath = "dashMantineComponentsFunctions") {
+    // Use the default path if the provided desc does not contain a dot
+    const path = desc.includes(".") ? desc : `${defaultPath}.${desc}`;
+
+    // allows for nested paths  for example "myNamespace.mySubNamespace.myFunction"
+    const arr = path.split(".");
+    while (arr.length && (obj = obj[arr.shift()]));
+    return obj;
+}
+
+function resolveProps(props, functionalProps, context){
+    let nProps = Object.assign({}, props);
+    for(let prop of functionalProps) {
+        if (nProps[prop]) {
+            nProps[prop] = resolveProp(nProps[prop], context);
+        }
+    }
+    return nProps
+}
+
--- a/src/ts/components/core/slider/Slider.tsx
+++ b/src/ts/components/core/slider/Slider.tsx
@@ -11,6 +11,7 @@ import { StylesApiProps } from "props/styles";
 import { TransitionProps } from "props/transition";
 import React, { useState } from "react";
 import { setPersistence, getLoadingState } from "../../../utils/dash3";
+import { resolveProp } from "../../../utils/prop-functions"
 
 interface Props
     extends BoxProps,
@@ -70,6 +71,7 @@ const Slider = ({
     loading_state,
     updatemode = "mouseup",
     value,
+    label,
     persistence,
     persisted_props,
     persistence_type,
@@ -94,6 +96,7 @@ const Slider = ({
             data-dash-is-loading={getLoadingState(loading_state) || undefined}
             {...others}
             value={val}
+            label={resolveProp(label)}
             onChange={setVal}
             onChangeEnd={(value) => {
                 if (updatemode === "mouseup") {


AnnMarieW:
Here is an example for the `renderOption` prop in the `Select`

```python


from dash import Dash

app = Dash()

    dmc.Select(
      label="Select with renderOption",
      placeholder="Select text align",
      data=[
        { "value": 'left', "label": 'Left' },
        { "value": 'center', "label": 'Center' },
        { "value": 'right', "label": 'Right' },
        { "value": 'justify', "label": 'Justify' },
      ],
      renderOption={"function": "renderSelectOption"}
    )

])

if __name__ == "__main__":
    app.run(debug=True)
```

```js
var dmcfuncs = window.dashMantineFunctions = window.dashMantineFunctions || {};

const icons = {
    left: React.createElement(window.dash_iconify.DashIconify, { icon: "mdi:format-align-left", width: 24 }),
    center: React.createElement(window.dash_iconify.DashIconify, { icon: "mdi:format-align-center", width: 24 }),
    right: React.createElement(window.dash_iconify.DashIconify, { icon: "mdi:format-align-right", width: 24 }),
    justify: React.createElement(window.dash_iconify.DashIconify, { icon: "mdi:format-align-justify", width: 24 })
};
const checkedIcon = React.createElement(window.dash_iconify.DashIconify, { icon: "mdi:check", width: 24 })

dmcfuncs.renderSelectOption=  function ({option, checked}) {
    return React.createElement(
        window.dash_mantine_components.Group,
        { flex: "1", gap: "xs" },
        icons[option.value],
        option.label,
        checked && checkedIcon
    );
}


```
image

AnnMarieW:
Here is an example of a `Select` component that has a custom filter.

See it live on [PyCafe](https://py.cafe/snippet/dash/v1?#c=H4sIAPCwFmgEA8VWbW_bNhD-K4RazHJi0295aVxoHdYtG4YNCNoA_RClBi3REROJ1Egqjuv6v-85SU5sIF2KfZkCxLq7596PR62DxKQymAaxVkVprGepcNmsENorLWeJAVNL7R0TjqVFEuuFNUUNYq3CL3iPdaxFWbKopsJuS_NcrEzlwYYq_6sxemHNvUqlDa9izfA0_wnwUeYy8WHDoCcVXkQtjJ7OH-JefEysKn2nt8O-XJXyGfbFymdG77E-VG4fQxb3GVWu9jkfM5nn7Dn7ZiltLd1j_2b2yPeHh_v0q86Wut7h52Iu8ygOLqXz7JPK00TYlJ2r3EsbBzvAVLo6FGU04OfGMvkgijKXrDxgzguLZi2Vz1jJGTtosEzqtOU2jD2LTgqbZGKey-jSVnJHsqjdR-s4WFQ6IZdxMGVxYGr3bhvdptGo235d_1cLNptpUcjZjEURNGaYKaVnM-g3YBoPW-kwlfPqpvYLxaAXWPl3pawsaOgwlzRpsf7WUP7EMu9LNx0MyhVPxEIObrIBKqAWIvEDp2Wm8vvbAan3W_X-k_pgMjw-ORqPj89GNeQZD_0RH_Nhv1xN-hqcvtArvsxyxOkxc218oMp61ECXK5Niuvv3Qz4-4SOImjMQTNfBvbQOZQumY6RpjP9glsRuLVlQvSBBxKmVAF09SryYO-khXKrUZ8F0dDzsBYXSnxpy0lC_S3WTwQ-RKoUamid_hlUn7XujPcov7Tc8ELQ_b7CAlILsBsHmetN7xDxG8ZwBCJ_0qLXlKtjX3RE72HF1xduVcN4Ol_svOvwWaruBfkchXy5eW4yXCret2V4Aj26y1uro9Ph7fOIsgS3yl5xuceSV_ja9uoE4LsDQoQOmLRiXRelX5ybHwr3IRSKz-pUMwwkmHNC66Ptq36rzrlqs74WlvU2rwWHHL5VOzZI_p_uC9OtXtt68pb0RazjAAlwia1qAl-aDvJEP0A8xCUhcd1n0I1vTDmmQ2IailCkQLYBbWVKm4eCqP4jxfH59-I6H3a9XcXy93lwPbnqsQ_zXP3S6cLo1ZMnRRWNjx5rSSV5h54adgw42FK2ud6zzucMOt66fHMbxQW2dA8oGg8csWOVk2uhOt1pvCaAN-8yETjJje6wQPslArZaZtJLgVvoKwWi5ZCjDrw9luBskHKk6g6Z0207wveWMTLarm4Vr1sp67c5nm-5uLducqV2NnHvzJ11y74WTYZd7qwr8uDJXPuywTpcXogy3Wba9ooh2gm89ckwobIdhQz91ceu7vv_gmGbcLFo13B0R6zj41Tcd1L3lTtsXfF9ACf6ezCyNxU0XNeb-LXw0yx0OmgF4DHZbAC6xq1dhU-461toud6aQYUjvNbMWc48bu-F1G3Mb-qGm4IBi_5vkjo4nnerm5LxiGU7BlL2Rw2S8OBsl8mQxeTOZz8fj07PTRAzF8Fikp2exnudK30kbRSN-xoc4HJLuNhVF4-H4CDfTGKxMWJxZXE-2ELn6QugJn9SiXCV3UfSGj_gp5gM-STTkR7H-f65NfD3mwiEkimKCr4RUCyIQXfMxmat5v5Be1B9-CPyUklbepULfSGsqh8zJcqxvlb4VY1IekaVC2LuqdEiFECN-HGst7DITOVSAOOUjcNCovnArnShD3BMyVObG56soAkGY7f1NBYoiXOLEpG8S6NbeJyPSwrzYFYayNk4FRQuq0htD_k7O-DGpOfVA8lHtB3MNfF8-YAjoKwC4Iz6qjVU2R-KTJjfoYWLvvkh8FdV1Oon1F1UiGPS0hjuHQgabfwCf4XxzuwsAAA)


```python

from dash import Dash

app = Dash()

    dmc.Select(
        data=[
        'JavaScript',
        'TypeScript',
        'Python',
        'Rust',
        'Java',
        'Julia',
        'Shell Script',
        'PowerShell',
        'Go',
        'C++',
        'C#'
      ],
        label="Test Wildcard Filter",
        description="For example p* starts with p.  *script ends with script",
        searchable=True,
        filter={"function": "optionsFilter"}
    )

])

if __name__ == "__main__":
    app.run(debug=True)
```

```js

var dmcfuncs = window.dashMantineFunctions = window.dashMantineFunctions || {};


const wildcardToRegex = (pattern) => {
  const escaped = pattern.replace(/[-/\\^$+?.()|[\]{}]/g, '\\$&');
  const regexPattern = pattern.includes('*')
    ? '^' + escaped.replace(/\*/g, '.*') // wildcard used
    : escaped; // no ^ anchor, match anywhere
  return new RegExp(regexPattern, 'i');
};

dmcfuncs.optionsFilter = function ({ options, search }) {
  const patterns = search.toLowerCase().trim().split(' ').map(wildcardToRegex);

  return options.filter((option) => {
    const label = typeof option === 'string' ? option : option.label;
    const words = label.toLowerCase().trim().split(/\s+/);
    return patterns.every((regex) => words.some((word) => regex.test(word)));
  });
};


```

AnnMarieW:
Here is an example of a nested function,  Here a function that creates a custom tooltip is passed to  Recharts using the `tooltipProps` property.

```python
tooltipProps={"content":  {"function": "customChartTooltip"}}
```



```python
from dash import Dash

app = Dash()
data = [
    {"month": "January", "Smartphones": 1200, "Laptops": 900, "Tablets": 200},
    {"month": "February", "Smartphones": 1900, "Laptops": 1200, "Tablets": 400},
    {"month": "March", "Smartphones": 400, "Laptops": 1000, "Tablets": 200},
    {"month": "April", "Smartphones": 1000, "Laptops": 200, "Tablets": 800},
    {"month": "May", "Smartphones": 800, "Laptops": 1400, "Tablets": 1200},
    {"month": "June", "Smartphones": 750, "Laptops": 600, "Tablets": 1000}
]

component = dmc.BarChart(
    h=300,
    dataKey="month",
    data=data,
    series=[
        {"name": "Smartphones", "color": "violet.6"},
        {"name": "Laptops", "color": "blue.6"},
        {"name": "Tablets", "color": "teal.6"}
    ],
    tickLine="y",
    gridAxis="y",
    withXAxis=True,
    withYAxis=True,
    valueFormatter={"function": "chartFormatter"},
    tooltipProps={"content":  {"function": "customChartTooltip"}}
)

    component
)

if __name__ == "__main__":
    app.run(debug=True)
    
```
    
```js
var dmcfuncs = window.dashMantineFunctions = window.dashMantineFunctions || {};

// bar chart value formatter

dmcfuncs.chartFormatter = (value) => new Intl.NumberFormat('en-US').format(value)


// bar chart custom tooltip

function ChartTooltip({ label, payload }) {
  if (!payload || payload.length === 0) return null;

  const dmc = window.dash_mantine_components;

  return dmc.Paper({
    px: "md",
    py: "sm",
    withBorder: true,
    shadow: "md",
    radius: "md",
    children: [
      dmc.Text({
        fw: 500,
        mb: 5,
        children: label,
      }),
      ...payload.map((item) =>
        dmc.Text({
          key: item.name,
          c: item.color,
          fz: "sm",
          children: `${item.name}: ${item.value}`,
        })
      ),
    ],
  });
}

dmcfuncs.customChartTooltip = ({ label, payload }) => ChartTooltip({ label, payload });


```

AnnMarieW:
@alexcjohnson This PR is still in draft but I'd love to get your input on what I have so far before applying it to all the components.  Do you think it's OK to start with only allowing functions defined in assets?

src/ts/utils/prop-functions.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+    // Check if the prop should be resolved as an arrow function.
+    if (prop.arrow){
+        return (...args) => prop.arrow
+    }
+    // If none of the special properties are present, do nothing.
+    return prop
+}
+
+function resolveVariable(prop, context){
```
omarirfa, line 38: The current resolveVariable function is limited, it can only pass the dynamic arguments when the function is called, plus the context:

```ts
if(isFunction(variable) && context){
    return (...args) => variable(...args, context)
}
```
This forces developers to create separate functions for each use case:

```js
// Current approach - duplicate functions for each variant
window.dashMantineFunctions.celsiusFormatter = (value) => `${value}°C`;
window.dashMantineFunctions.fahrenheitFormatter = (value) => `${value}°F`;
window.dashMantineFunctions.kelvinFormatter = (value) => `${value}K`;

// Need separate functions for different currencies too
window.dashMantineFunctions.usdFormatter = (value) => `$${value.toFixed(2)}`;
window.dashMantineFunctions.eurFormatter = (value) => `€${value.toFixed(2)}`;
```

Modifying the resolveVariable function slightly can help it accept arguments:
```ts
if(isFunction(variable) && context){
    // New code - support for args
    if(prop.args) {
        return (...callArgs) => variable(...(Array.isArray(prop.args) ? prop.args : [prop.args]), ...callArgs, context)
    }
    return (...args) => variable(...args, context)
}
```

Here is an example of how the new way with reusable functions will look like:
```js
// js file in assets folder
// Create one function that works for multiple cases
window.dashMantineFunctions.formatTemperature = (unit, value) => `${value}°${unit}`;
window.dashMantineFunctions.formatCurrency = (currency, value) => {
    const symbols = {"USD": "$", "EUR": "€", "GBP": "£"};
    return `${symbols[currency] || ""}${value.toFixed(2)}`;
};
```

```python
# Python usage - same function, different args
dmc.Slider(value=25, label={"function": "formatTemperature", "args": "C"})
dmc.Slider(value=25, label={"function": "formatTemperature", "args": "F"})

# Multiple args as array
dmc.NumberInput(
    value=100, 
    formatter={"function": "formatCurrency", "args": "USD"},
    parser={"function": "parseCurrency", "args": ["USD", 2]}
)
```

Impact of adding function arguments:
- [x] **Eliminate code duplication** - Write one function that handles multiple cases instead of duplicating nearly identical functions
- [x] **Centralized Logic** - Fix bugs in one place instead of hunting down multiple functions
- [x] **Better Python Readability** - `{"function": "formatTemp", "args": "C"}` clearly shows intent
- [x] **Reduced Maintenance** - Update one function instead of 10+ variants when requirements change

would love some thoughts on if this seems feasible!

omarirfa:
Some other ideas, I have not had a chance to flesh them out, so I am thinking out loud here:

1. Maybe add some error handling to say if a function is missing/misnamed in resolveVariable and perhaps would it be worthwhile adding  a fallback? for example the function errors out since function name is incorrect, instead of erroring out it can resort to a fallback if defined by the user?
2.  It would be nice to have the option to memoize components especially if you have a large data set and for example you open a select menu showing 1000 items. Every time a component renders and since DMC components render frequently, the code will recreate the function and there will be a loss of referential equality.

alexcjohnson:
@AnnMarieW I like it! Personally I'd be inclined to stay with only functions defined in assets until there's a real clear need to accept functions as strings.

@omarirfa lots of great ideas, thank you! Before we add any extra error handling we should see how the current implementation meshes with dash devtools - if that makes the error clear I'd be inclined to not implement a fallback, better to be explicit and prompt the dev to fix it. And re: memoization - this could certainly be important in some use cases, but we'll need to be careful about it so that we avoid both improperly returning a cached value and allowing the cache to explode in size. Let's get this merged and then make a followup PR to add this once we've collected a few use cases to test it against.

AnnMarieW:
Thanks for you both for your feedback!
@alexcjohnson What do you think about @omarirfa 's idea about passing args from the dash app to the js function?  That would add a lot of nice functionality, but would it introduce vulnerabilities?

If the function can't be found or if there is a syntax error, the following message will show in dev tools:

image

Could probably improve the message, but id does work nicely with the dash devtools.

src/ts/utils/prop-functions.ts
alexcjohnson: This is a cool idea! For extra args like this a very common JavaScript convention is to use an "options" object that comes last in the args list, so you only need one extra argument. Again though I think we can do this in a followup PR after the base functionality is merged.

alexcjohnson: To @AnnMarieW 's query "would it introduce vulnerabilities?" I think it's safe. It doesn't allow arbitrary code execution, just arbitrary data as arguments.

src/ts/utils/prop-functions.ts
```diff
@@ -0,0 +1,70 @@
...
+    const variable = getDescendantProp(window, prop.function)
+
+    // If it's not there, raise an error.
+    if(variable === undefined){
+        throw new Error("No match for [" + prop.function + "] in the global window object.")
```
alexcjohnson, line 43: Nice! Can we say "in window.dashMantineFunctions"?

AnnMarieW: Yes that's the default.  But it also allows you to create other namespaces.  You can use them like this:
{"function": "myNamespce.myFunction"

AnnMarieW: I'd also like to improve the message to include something about syntax errors.  If there are any errors in the namespace then you will see this error message, which is a little confusing.

alexcjohnson: > But it also allows you to create other namespaces.

Oh I missed that - seems a little risky actually to let people access any code on `window`, also a little confusing in that if you want something directly inside `dashMantineFunctions` you can omit the namespace but if you want something nested inside `dashMantineFunctions` you need to include it. I'd be more comfortable if this path always started at `dashMantineFunctions`.

src/ts/utils/prop-functions.ts
```diff
@@ -0,0 +1,70 @@
...
+        return prop
+    }
+    // Check if the prop should be resolved a variable.
+    if (prop.function){
+        return resolveVariable(prop, context)
+    }
+    // Check if the prop should be resolved as an arrow function.
+    if (prop.arrow){
+        return (...args) => prop.arrow
+    }
```
alexcjohnson, line 33: Seems like this is actually resolving as a constant, isn't it? Should we call this `prop.constant`?

AnnMarieW: Yeah, this is a placeholder for now.  It might be better to just take it out.  (This is all based on dash-extensions.js and it's how dash-leaflet passes functions as props.)

AnnMarieW:
Added the ability to pass optional args to the function:
```python
{"function": "labelFormatter", "options": {"suffix": " °F"}},
```


```python
from dash import Dash
app = Dash()

    dmc.Slider(value=25, p=50, label={"function": "labelFormatter", "options": {"suffix": " °F"}}),

])

if __name__ == "__main__":
    app.run(debug=True)
```

```js
var dmcfuncs = window.dashMantineFunctions = window.dashMantineFunctions || {};

dmcfuncs.labelFormatter = function (value, opts = {}) {
  const prefix = opts.prefix || '';
  const suffix = opts.suffix || '';
  return `${prefix}${value}${suffix}`;
};
```

image


        
"""

omarirfa:
thoughts on adding type safety? I just assumed the types here in the sample implementation for prop-functions.ts

```ts
/**
 * Utility to allow passing a function as a prop from a dash app.
 * For example label={'function': 'myLabel'}
 * where 'myLabel' is a function defined in .js file in /assets
 */

// Define the structure of a function prop
interface FunctionProp {
  function: string;
  options?: Record<string, any>;
  args?: any[] | any;
}

// Define the structure of possible context objects
interface PropContext {
  [key: string]: any;
}

// Type guard to check if an object is a plain object
function isPlainObject(o: any): boolean {
  return !(o === null || o === undefined || 
           Array.isArray(o) || 
           typeof o === 'function' || 
           o.constructor === Date) && 
          (typeof o === 'object');
}

// Type guard to check if a value is a function
function isFunction(functionToCheck: any): functionToCheck is Function {
  return Boolean(functionToCheck) && 
         {}.toString.call(functionToCheck) === '[object Function]';
}

// Type guard to check if an object is a FunctionProp
function isFunctionProp(prop: any): prop is FunctionProp {
  return isPlainObject(prop) && 
         typeof prop.function === 'string';
}

export function resolveProp<T = any>(prop: any, context: PropContext = {}): T {
  // If it's not an object, just return
  if (!isPlainObject(prop)) {
    return prop as T;
  }
  
  // Check if the prop should be resolved as a variable
  if (isFunctionProp(prop)) {
    return resolveVariable(prop, context) as T;
  }
  
  // If none of the special properties are present, do nothing
  return prop as T;
}

function resolveVariable(prop: FunctionProp, context: PropContext): any {
  const options = prop.options || {};
  const variable = getDescendantProp(window, prop.function);

  // If it's not there, raise an error
  if (variable === undefined) {
    throw new Error(`No match for [${prop.function}] in window.dashMantineFunctions.`);
  }
  
  // If it's a function, add context
  if (isFunction(variable)) {
    if (prop.args) {
      return (...callArgs: any[]) => 
        variable(...(Array.isArray(prop.args) ? prop.args : [prop.args]), 
                 ...callArgs, options, context);
    }
    return (...args: any[]) => variable(...args, options, context);
  }
  
  // Otherwise, use the variable as-is
  return variable;
}

function getDescendantProp(obj: any, name: string): any {
  if (name.includes(".")) {
    throw new Error(
      `Function name "${name}" is invalid. Dotted paths are not allowed. Only functions in the "dashMantineFunctions" namespace are supported.`
    );
  }

  const ns = obj.dashMantineFunctions;
  return ns ? ns[name] : undefined;
}

export function resolveProps<T extends Record<string, any> = Record<string, any>>(
  props: T | null | undefined, 
  context: PropContext = {}
): T {
  if (!props || typeof props !== 'object') return {} as T;

  return Object.fromEntries(
    Object.entries(props).map(([key, value]) => [key, resolveProp(value, context)])
  ) as T;
}

```

Testing

1. All functions still behave exactly the same at runtime
2. Added proper type safety for compile-time checking

omarirfa:
A few other ideas I was thinking, even if they are added later on, just writing somewhere before i forget 

1. Since we have function arguments, thought on having function composition i.e. chaining multiple functions and passing it ?
2. The user defined functions can be validated and verified by the script as well in a very limited scope ? 
3. Adding some common utility function I think will be a nice start for users to use this!

AnnMarieW:
@omarirfa Thanks for adding the type checking.  However, it doesn't seem to check the types - for example I was able to have a prop like this {"function": True}.  It may not throw an error because this function is checking for a string:

```

// Type guard to check if an object is a FunctionProp
function isFunctionProp(prop: any): prop is FunctionProp {
  return isPlainObject(prop) && 
         typeof prop.function === 'string';
}

``` 

It would be better to either throw an error if it's not a string, or enable the type checker to handle it.

omarirfa:
@AnnMarieW Ah you are right, I forgot to add the errors. Heres an updated version of the code which addresses your concern and some other test/edge cases:

```typescript
/**
 * Utility to allow passing a function as a prop from a dash app.
 * For example label={'function': 'myLabel'}
 * where 'myLabel' is a function defined in .js file in /assets
 *
 * Based on https://github.com/emilhe/dash-extensions-js
 */

// Define interfaces for type safety
export interface FunctionProp {
  function: string;
  options?: Record<string, any>;
}

export type DashProp = FunctionProp | any;

// Helper to detect circular references
function hasCircularReference(obj: any, seen = new WeakSet()): boolean {
  if (obj === null || typeof obj !== 'object') return false;
  if (seen.has(obj)) return true;
  
  seen.add(obj);
  for (const key of Object.keys(obj)) {
    if (hasCircularReference(obj[key], seen)) return true;
  }
  return false;
}

function isPlainObject(o: any): boolean {
  return (o === null || o === undefined || Array.isArray(o) || typeof o === 'function' || o.constructor === Date) ?
    false : (typeof o === 'object');
}

function isFunction(functionToCheck: any): functionToCheck is Function {
  return functionToCheck && {}.toString.call(functionToCheck) === '[object Function]';
}

function isFunctionProp(prop: any): prop is FunctionProp {
  if (!isPlainObject(prop)) {
    return false;
  }
  
  // Check if it has a 'function' property that is a string
  if ('function' in prop) {
    if (typeof prop.function !== 'string') {
      throw new Error(`Invalid function prop: 'function' property must be a string, got ${typeof prop.function}`);
    }
    
    // Check for empty function name
    if (prop.function.trim() === '') {
      throw new Error(`Invalid function prop: 'function' property cannot be an empty string`);
    }
  } else {
    return false;
  }
  
  // If options is provided, check that it's an object
  if ('options' in prop && prop.options !== undefined) {
    if (!isPlainObject(prop.options)) {
      throw new Error(`Invalid function prop: 'options' property must be an object, got ${typeof prop.options}`);
    }
    
    // Check for circular references in options
    if (hasCircularReference(prop.options)) {
      throw new Error(`Invalid function prop: 'options' contains circular references`);
    }
  }
  
  return true;
}

export function resolveProp(prop: DashProp, context: Record<string, any> = {}): any {
  // Handle null and undefined explicitly
  if (prop === null || prop === undefined) {
    return prop;
  }
  
  // If it's not an object, just return
  if (!isPlainObject(prop)) {
    return prop;
  }
  
  try {
    // Check if the prop should be resolved as a function prop
    if ('function' in prop) {
      if (isFunctionProp(prop)) {
        return resolveVariable(prop, context);
      }
    }
  } catch (error) {
    console.error("Error in resolveProp:", error);
    throw error; // Re-throw to make type errors visible
  }
  
  // If none of the special properties are present, do nothing
  return prop;
}

function resolveVariable(prop: FunctionProp, context: Record<string, any> = {}): any {
  if (!prop.function) {
    throw new Error("Function name is required");
  }
  
  const options = prop.options || {};
  
  try {
    // Check if dashMantineFunctions exists
    if (!window.dashMantineFunctions) {
      throw new Error("window.dashMantineFunctions is not defined");
    }
    
    const variable = getDescendantProp(window, prop.function);

    // If it's not there, raise an error
    if (variable === undefined) {
      throw new Error(`No match for [${prop.function}] in window.dashMantineFunctions.`);
    }
    
    // If it was found but isn't a function, warn but return it anyway
    if (!isFunction(variable)) {
      console.warn(`Warning: [${prop.function}] is not a function, it's a ${typeof variable}`);
      return variable;
    }
    
    // If it's a function, add context
    if (context && Object.keys(context).length > 0) {
      return (...args: any[]) => variable(...args, options, context);
    } else {
      return (...args: any[]) => variable(...args, options);
    }
  } catch (error) {
    console.error("Error in resolveVariable:", error);
    throw error;
  }
}

function getDescendantProp(obj: any, desc: string): any {
  // Ensure all lookups stay within dashMantineFunctions
  const root = obj.dashMantineFunctions;
  if (!root) {
    throw new Error("dashMantineFunctions is not defined in the window object");
  }
  
  if (typeof desc !== "string") {
    throw new Error(`Property path must be a string, got ${typeof desc}`);
  }
  
  if (desc.trim() === "") {
    throw new Error("Property path cannot be empty");
  }
  
  const path = desc.split(".");
  let current = root;

  for (const segment of path) {
    if (segment.trim() === "") {
      throw new Error(`Invalid path segment: empty segment in "${desc}"`);
    }
    
    if (!Object.prototype.hasOwnProperty.call(current, segment)) {
      return undefined;
    }
    current = current[segment];
  }

  return current;
}

export function resolveProps(props: Record<string, DashProp> | null | undefined, context: Record<string, any> = {}): Record<string, any> {
  if (props === null || props === undefined) {
    return {};
  }
  
  if (typeof props !== 'object' || Array.isArray(props)) {
    throw new Error(`Props must be an object, got ${typeof props}`);
  }
  
  try {
    return Object.fromEntries(
      Object.entries(props).map(([key, value]) => {
        try {
          return [key, resolveProp(value, context)];
        } catch (error) {
          console.error(`Error resolving prop "${key}":`, error);
          throw new Error(`Error resolving prop "${key}": ${error.message}`);
        }
      })
    );
  } catch (error) {
    console.error("Error in resolveProps:", error);
    throw error;
  }
}

```


Improvements:

1. Properly rejects invalid function types (e.g., {"function": true})
2. Handles null/undefined values in props
3. Handles empty strings and whitespace-only function names
4. Handles circular references in options objects
5. Handles invalid path formats and empty path segments
6. Handles non-object options
7. Handles missing window components
8. Handles non-function values in dashMantineFunctions


Error message output:

1. Non string function value
`ERROR: Invalid function prop: 'function' property must be a string, got boolean`
2. Non-object options
`ERROR: Invalid function prop: 'options' property must be an object, got string`
3. Empty function name
` ERROR: Invalid function prop: 'function' property cannot be an empty string `
4. Whitespace function name
` ERROR: Invalid function prop: 'function' property cannot be an empty string  `
5. Circular reference in options
`ERROR: Invalid function prop: 'options' contains circular references `
6. Invalid function path
`ERROR: No match for [nonexistent.function] in window.dashMantineFunctions. `
7. Invalid path format
`ERROR: No match for [nested..function] in window.dashMantineFunctions. `
9. resolveProps with non-object
` ERROR: Props must be an object, got string`
10. resolveProps with invalid prop
` ERROR: Error resolving prop "invalid": Invalid function prop: 'function' property must be a string, got boolean `
11. Missing dashMantineFunctions
` ERROR: window.dashMantineFunctions is not defined in the window object `

alexcjohnson (approved):
💃 LGTM! You're still intending to remove [assets/dashMantineFunctions.js](https://github.com/snehilvj/dash-mantine-components/pull/580/files#diff-7018374b45c4edeff70d69466e007edc7438e61f9d583538bf1af5697daecdf5), right?

AnnMarieW:
Yes, I'll remove assets/dashMantineFunctions.js before merging. Just wanted to keep it there during review to help run examples locally in the PR.

Thanks for the review!  I'm excited about this new feature :slightly_smiling_face:

PR #584, opened 2025-05-08
DMC 2.0 with Mantine 8.0

### LINKED ISSUE #583: [regression] "format" prop is gone from TimeInput
I just tried upgrading DMC in my environment and am getting this error when running my Dash app
```
Exception has occurred: TypeError
The `dash_mantine_components.TimeInput` component (version 1.1.1) with the ID "start-time--" received an unexpected keyword argument: `format`
```
Before upgrading the same code worked fine.
The code being like
```
return  dmc.TimeInput(
        label=label,
        id=id_,
        withSeconds=True,
        format="24",
        value=datetime.datetime.now(),
        style={"width": 180},
    )
```
As you can see, I set the `format` prop to `"24"` to use the 24 hour format instead of the AM/PM one. But this property got lost at some point.
This is important as people in my organization specifically requested the 24 hour format and we use it all over the place.

AnnMarieW:
Mantine V8 Changelog:  https://mantine.dev/changelog/8-0-0/
Migration guide: https://mantine.dev/guides/7x-to-8x/

Date components
- [x]  Updated date components to use strings instead of date objects
- [x] Add `timePickerProps` to `DateTimePicker`  (changed prop name from `timeInputProps`)
- [x] Remove `preserveTime` prop from `DateInput`
- [x] Added new `TimePicker` component closes #583 

Carousel
- [x]  Added `emblaOptions`
- [x] Removed props `loop, align, slidesToScroll, dragFree, inViewThreshold, skipSnaps, containScroll` 

Switch
- [x] add `withThumbIndicator` prop

CodeHighlight
Note that including only the top 10 languages instead of all (which was the default in Mantine 7) the bundle size for the CodeHighlight components went from 948K to 85.8K :tada: 
- [x] bundle highlight.js
- [x] add `CodeHighlightAdapterProvider`  
- [x] bundle default highlight.js theme 

Popover
- [x]  add `hideDetatched` prop

Kbd
- [x]  add `size`  prop (already there)

Docs only  
- [x] Write migration guide
- [x] `Portal` changed default for the `reuseTargetNode` 
- [x] `DatesProvider`  no longer supports `timezone` option
- [x] `Menu data-hovered` attribute
- [x] `Popover` `hideDetatched` changed default 
- [x] `ScrollArea`  add `content` Styles API selector 
- [x] `Image` style changes and new examples
- [x] Check the "other changes" section of Mantine V8 Changelog
- [x] CodeHighlight - document how to add extra languages and how to add external stylesheets from highlight.js

New components added in this PR
- [x]  `TimePicker`
- [x]  `TimeGrid`  
- [x]   `Menu` with sub menus
- [x] `DatePicker` (calendar picker without an input field)

New components to add in a separate PR (If necessary)
-  `Heatmap`   May not add this since it's possible to make a heatmap like this using Plotly
-  `TimeValue`  May not need this one.  Value can be formatted in Python and displayed as as string.  

Draft PR for the V2 docs: https://github.com/snehilvj/dmc-docs/pull/194


src/ts/components/core/menu/MenuSub.tsx
```diff
@@ -0,0 +1,65 @@
...
+    /** Set the `tabindex` on all menu items. Defaults to -1 */
+    menuItemTabIndex?: -1 | 0;
+    /** Unique ID to identify this component in Dash callbacks. */
+    id?: string;
+    /** Update props to trigger callbacks. */
+    setProps: (props: Record<string, any>) => void;
+}
+
+/** Sub Menu */
+const MenuSub = (props: Props) => {
```
alexcjohnson, line 42: Should it be called `SubMenu` instead?

alexcjohnson: Not a real strong opinion but I don't think it would be too big a stretch to let `Menu.Sub` translate to `SubMenu`. Some namespaced components make sense in the original order, some not so much. I feel like just pointing this out in the docstring would suffice, like "`SubMenu` is a Mantine `Menu.Sub` component."

AnnMarieW: Yeah, I kinda like that better too.

PR #609, opened 2025-06-27
8.1.2 update

### LINKED ISSUE #548: MultiSelect fails to render when default value is None
I'm migrating to DMC 2.0 and this bug took a while to hunt down. From the documentation, it looks like the default value of MultiSelect has to be a list of strings. However, if the default value is not a list (e.g., None), the MultiSelect doesn't render and there is no useful error in the debug or python console. The only error that pops up is `Re is null` in the debug panel. It would be nice if this gave a TypeError or something if the default value isn't a list of strings: 

```
from dash import Output, Input, html, callback
import dash

app = dash.Dash(__name__)


app.layout = dmc.MantineProvider(html.Div(
    [
        dmc.MultiSelect(
            label="Select your favorite libraries",
            placeholder="Select all you like!",
            id="framework-multi-select",
            value=None,
            data=[
                {"value": "pd", "label": "Pandas"},
                {"value": "np", "label": "NumPy"},
                {"value": "tf", "label": "TensorFlow"},
                {"value": "torch", "label": "PyTorch"},
            ],
            w=400,
            mb=10,
        ),
        dmc.Text(id="multi-selected-value"),
    ]
))

@callback(
    Output("multi-selected-value", "children"), Input("framework-multi-select", "value")
)
def select_value(value):
    if value is None:
        return "nothing selected"
    return ", ".join(value)

app.run(debug=True)
```

AnnMarieW:
Updated to Mantine 8.1.2:
- added `headerControlOrderProp` to calendar components (added example in docs)
- added `presets` to `DatePicker` DateTimePicker` and  (undocumented because I think callbacks work better in Dash)
- added `autoContrast` to Tooltip and FloatingTooltip
- added `domain` to `Slider` and `RangeSlider`  (added example in docs)
- added `pushOnOverlap` to `RangeSlider (added example in docs)
- added `bdrs` new style prop (updated docs)

Not new in Mantine 8.1, but added here to make available in the next release:
- added `getYearControlProps`, `getMonthControlProps`, `getDayProps`, `renderDay` to date components

docs PR https://github.com/snehilvj/dmc-docs/pull/215

No tests needed since they are all pass-through props

closes #548 
 - handles MultiSelect when `value=None`
 
Todo
- [x] changelog

### DIFF
--- a/src/ts/components/core/combobox/Autocomplete.tsx
+++ b/src/ts/components/core/combobox/Autocomplete.tsx
@@ -12,7 +12,7 @@ import { ScrollAreaProps } from 'props/scrollarea';
 import { StylesApiProps } from 'props/styles';
 import React, { useState } from 'react';
 import { setPersistence, getLoadingState } from '../../../utils/dash3';
-import { parseFuncProps } from "../../../utils/prop-functions"
+import { parseFuncProps } from '../../../utils/prop-functions';
 
 interface Props
     extends BoxProps,
--- a/src/ts/components/core/slider/RangeSlider.tsx
+++ b/src/ts/components/core/slider/RangeSlider.tsx
@@ -74,7 +74,7 @@ interface Props
     /** Domain of the slider, defines the full range of possible values, `[min, max]` by default */
     domain?: [number, number];
     /** Determines whether the other thumb should be pushed by the current thumb dragging when minRange/maxRange is reached, true by default */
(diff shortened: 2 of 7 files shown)


alexcjohnson (approved):
Looks good, maybe add a `### Fixed` in the changelog for the CheckboxGroup` and `MultiSelect` fixes?

PR #616, opened 2025-07-09
Fix Slider and RangeSlider when min or max changes in a callback

### LINKED ISSUE #615: [bug] Slider and RangeSlider break if range is changed in a callback.
he slider breaks after changing its range in a callback and dragging it, same with the range slider. Here is my code (dash 2.15.0, dmc 2.1.0):

```python
from dash import Dash, Output, Input, State, callback, _dash_renderer


app = Dash()

layout = dmc.Box(
    children = [
        dmc.Text("Normal slider"),
        dmc.Slider(
            min = 0,
            max = 100,
            value = 100,
            step = 1,
        ),
        dmc.Text("Bug"),
        dmc.Button(id = "add_100", children = "Add 100 to max"),
        dmc.Slider(
            id = "bug_slider",
            min = 0,
            max = 100,
            value = 100,
            step = 1,
        ),
    ],
    p = "md"
)

app.layout = dmc.MantineProvider(layout)

@callback(
    Output("bug_slider", "max"),
    Input("add_100", "n_clicks"),
    State("bug_slider", "max"),
    prevent_initial_call = True
)
def add(n_clicks, max):
    return max + 100


if __name__ == '__main__':
    app.run(debug=True)


```

AnnMarieW:
Fixes #615 

Dragging the handle causes `MantineSlider` to calculate the new value based on the handle’s position. However,  the `MantineSlider` is storing the drag’s position as a percent of the old range.   

Using  a key based on the `min` and `max` remounts the slider whenever min or max changes, resetting the drag state.

### DIFF
--- a/src/ts/components/core/slider/RangeSlider.tsx
+++ b/src/ts/components/core/slider/RangeSlider.tsx
@@ -102,6 +102,7 @@ const RangeSlider = ({
 
     return (
         <MantineRangeSlider
+            key={`${others.min}-${others.max}`}
             data-dash-is-loading={getLoadingState(loading_state) || undefined}
             {...parseFuncProps('RangeSlider', others)}
             value={val}
--- a/src/ts/components/core/slider/Slider.tsx
+++ b/src/ts/components/core/slider/Slider.tsx
@@ -94,6 +94,7 @@ const Slider = ({
 
     return (
         <MantineSlider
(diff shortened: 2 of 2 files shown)


alexcjohnson (approved):
💃 Good job diagnosing the issue, and very nice solution!

Issue #239, opened 2024-04-15
RadioGroup not default Horizontal aligned


slyin87:
Hi,

In the documentation, the default is aligned horizontally..

https://www.dash-mantine-components.com/components/radiogroup

but now it is going downwards. How to make it go horizontally?

version 0.13.0a3

slyin87:
TypeError: The `dash_mantine_components.RadioGroup` component (version 0.13.0a3) with the ID "chats_source_list" received an unexpected keyword argument: `orientation`

`Allowed arguments: aria-*, bg, bga, bgp, bgr, bgsz, bottom, c, children, className, classNames, data-*, description, descriptionProps, display, error, errorProps, ff, fs, fw, fz, h, id, inputWrapperOrder, inset, label, labelProps, left, lh, lts, m, mah, maw, mb, mih, miw, ml, mr, mt, mx, my, name, opacity, p, pb, persisted_props, persistence, persistence_type, pl, pos, pr, pt, px, py, required, right, size, style, styles, sx, ta, td, top, tt, unstyled, value, w, withAsterisk, wrapperProps`

datenzauberai:
It seems to me to work as documented. You either wrap the Radio components in a Group (left to right) or in a Stack (top to bottom). @slyin87 could you try using a Group that wraps your Radio components?

AnnMarieW:
Thanks for confirming @datenzauberai.
This also might  be an old issue since @slyin87 was running 13.0a3.

PR #625, opened 2025-08-11
Update to 8.2.4

AnnMarieW:
### Added
- `attributes` prop – Pass custom attributes to inner elements of components that support the Styles API.
- `strategy` prop for `Container` – Added `strategy="grid"` option for grid-based layouts with breakout features.
- `target` prop for `Tooltip` – Specify tooltip targets using a selector, HTML element, or ref, as an alternative to children.
- `chevronIconSize` prop for `Accordion` – Control the size of the default chevron icon independently from `chevronSize`.
- `keepMounted` prop for `Collapse` – Keep collapsed content in the DOM while hidden.
- `autoSelectOnBlur` prop for `Select` and `Autocomplete` – Automatically select the highlighted option when the input loses focus.

### Fixed
- Fixed CSS for `CodeHighlight` so it works when `dcc.Markdown` and/or `dash-ag-grid` are also used in the same app. 

### Changed
- Upgraded to latest Mantine (8.2.4) 

### Notes
- The one `Select` test that's failing is flakey  and can be disregarded (the test passes locally).
- no new tests are needed because all the new props are pass-through props only.  The docs have examples of each new feature and it all works as expected.   See  [Doc PR #239](https://github.com/snehilvj/dmc-docs/pull/239) 
-  This PR updates to Mantine 8.2.4 (the current version).  Will wait for 8.2.5 to fix a bug in `autoSelectOnBlur` before releaseing DMC 2.2.0.  Updating to 8.2.5 can be a separate PR


CHANGELOG.md
```diff
@@ -1,5 +1,26 @@
...
+- `keepMounted` prop for `Collapse` – Keep collapsed content in the DOM while hidden.
+- `autoSelectOnBlur` prop for `Select` and `Autocomplete` – Automatically select the highlighted option when the input loses focus.
+
+
+### Fixed
+- Fixed CSS for `CodeHighlight` so it works when `dcc.Markdown` and/or `dash-ag-grid` are also used in the same app. #625 by @AnnMarieW
+
+
+### Changed
+- Upgraded to latest Mantine (8.2.5) #625 by @AnnMarieW
```
alexcjohnson, line 22:
```suggestion
- Upgraded to latest Mantine (8.2.4) #625 by @AnnMarieW
```
?

alexcjohnson (approved):
💃 Looks good, just the version number mismatch changelog vs package.json

CHANGELOG.md
AnnMarieW: oh - nice catch.  My original plan was to wait for 8.2.5, but then thought it would be good to finish this one up now and do 8.2.5 in a separate PR but forgot to update the changelog.

alexcjohnson: OIC you're waiting for that before releasing... anyway best practice to keep the changelog and code in sync.

PR #627, opened 2025-08-13
add clearSearchOnFocus to Select

### LINKED ISSUE #626: [Feature Request] allow dmc.Select search to clear search text when dropdown is selected, similar to dcc.Dropdown
When I click a dmc.Select dropdown, the existing selected item text is in the text search bar. If I type a new search string, it is appended to that search text that is already there. If I want a fresh new search string, I have to first delete the existing text. This is not necessary with dcc.Dropdown. When I click a dcc.Dropdown, the search text cursor is reset to the beginning of the Dropdown box (even though the currently selected item is displayed in the search bar initially); When I type, it immediately clears the existing search text. This request is to allow dmc.Select to behave just like dcc.Dropdown in this regard. Perhaps add a new field to dcc.Select(), i.e. clearSearchValueOnOpen=True/False. It is fine for the currently selected dmc.Select item to show, but as soon as I type, I would like that pre-existing search string to be cleared. This allows for quicker dropdown selection searching in many instances.

AnnMarieW:
closes #626 

adds `clearSearchOnFocus` so a searchable Select is similar to the dcc.Dropdown


src/ts/components/core/combobox/Select.tsx
```diff
@@ -42,6 +42,10 @@ interface Props
     hiddenInputProps?: object;
     /** Props passed down to the underlying `ScrollArea` component in the dropdown */
     scrollAreaProps?: ScrollAreaProps;
+    /** If set, the highlighted option is selected when the input loses focus @default `false` */
+    autoSelectOnBlur?: boolean;
```
alexcjohnson, line 46: Is this one important enough to include a test for? That said when I try it on their [docs site](https://mantine.dev/core/select/#autoselectonblur) it doesn't seem to do what it says 🤷

AnnMarieW: Oh.  I just answered a different question. I thought you were asking about the `clearSearchOnFocus` which is the new feature added in this PR.

Their  `autoSelectOnBlur` is not currently working as they describe in their docs.  (It's a bug.)   I reported it and they said it will be fixed in  8.2.5.  I'll wait for that before releasing

alexcjohnson (approved):
💃

PR #637, opened 2025-08-23
update data and value together - fixes #634

### LINKED ISSUE #634: dmc.MultiSelect duplicate callback triggers from one action
I am running into unexpected behavior with dmc MultiSelect, where updating `data` prop in a way that also modifies `value`, will trigger 2 separate callbacks from a callback listening to both `data` and `value`. Using `dcc.Dropdown` will in the same situation trigger callback just once, with both inputs triggered.
It's easy to see with below example (python 3.13, dash 3.2.0, dmc 2.1.0):
```
from dash import Dash, callback, Output, Input, dcc, ctx


app = Dash()

    [
        dmc.Text(id="dmc-value"),
        dmc.Text(id="dcc-value"),
        dmc.MultiSelect(
            placeholder="dmc MultiSelect",
            id="dmc-dropdown",
            debounce=True,
        ),
        dcc.Dropdown(
            placeholder="dcc Dropdown",
            id="dcc-dropdown",
            multi=True,
        ),
        dmc.Button(
            "change options",
            id="change-options",
        ),
    ]
)


@callback(
    Output("dmc-dropdown", "data"),
    Output("dcc-dropdown", "options"),
    Input("change-options", "n_clicks"),
    prevent_initial_call=True,
)
def change_options(n_clicks):
    options = ["one", "two", "three"] if n_clicks % 2 == 0 else ["three", "four", "five"]
    print(n_clicks, "CHANGING OPTIONS")

    return options, options


@callback(
    Output("dmc-value", "children"),
    Input("dmc-dropdown", "value"),
    Input("dmc-dropdown", "data"),
)
def dmc_select_value(value, data):
    print(ctx.triggered)
    print(f"DMC VALUE: {value}, OPTIONS: {data}")
    return f"DMC VALUE: {value}, OPTIONS: {data}"


@callback(
    Output("dcc-value", "children"),
    Input("dcc-dropdown", "value"),
    Input("dcc-dropdown", "options"),
)
def dcc_select_value(value, options):
    print(ctx.triggered)
    print(f"DCC VALUE: {value}, OPTIONS: {options}")
    return f"DCC VALUE: {value}, OPTIONS: {options}"


if __name__ == "__main__":
    app.run(debug=True)
```
Click the button and observe log for `ctx.triggered`. On initial option load it works as expected, one callback, with both `data` and `value` triggered (as value actually changes from `None` to `[]`.
Next select some value in dropdown that are not present in the other option set. For example `three` and `four`. Now press the button again to change `options`/`data`. In both implementations, `value` will change to the overlap between previous selection and new options, in this case to `['three']`. However in the case of `dcc`, callback will be triggered only once, by both `options` and `value`, whereas with `dmc` it will trigger twice, once for each input.
I find this behavior unwanted, but am not sure if it's intended or not.

As a bonus, doing the same with `dmc.Select` will visually clear the selection when selected value is no longer available in `data`, but the `value` prop will not change.

AnnMarieW:
closes #634 

- fixes 'MultiSelect` and `Select` so that changes to the `data` and `value` are batched so they only trigger a single callback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- fixes 'MultiSelectandSelectso that changes to thedataandvalue` are batched so they only trigger a single callback. #637 by @AnnMarieW
+- fixes `MultiSelect` and `Select` so that changes to the `data` and `value` are batched so they only trigger a single callback. #637 by @AnnMarieW
 
 # 2.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+# Unreleased
+
+### Fixed
+
+- fixes 'MultiSelectandSelectso that changes to thedataandvalue` are batched so they only trigger a single callback. #637 by @AnnMarieW
+
 # 2.2.0
 
 See details and examples in the [DMC 2.2.0 Release Announcement](https://www.dash-mantine-components.com/release-2-2-0)
--- a/src/ts/components/core/combobox/MultiSelect.tsx
+++ b/src/ts/components/core/combobox/MultiSelect.tsx
@@ -85,20 +85,6 @@ const MultiSelect = ({
         }
     }, [debounced]);
 
-    useDidUpdate(() => {
-        const newOptions = Array.isArray(data) ? data : [];
-        const rawValue = Array.isArray(value) ? value : [];
-        const newSelected = filterSelected(newOptions, rawValue) ?? [];
-
-        setOptions(newOptions);
-        setSelected(newSelected);
-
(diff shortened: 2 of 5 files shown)


alexcjohnson:
Ah yes, that's better, good thinking!

Makes me realize though, it might be overkill but in the tests in principle just looking at `ctx.triggered` leaves open the possibility that the callback was executed more than once and this is just the last time, so to be precise (given that the bug is about unnecessary callbacks) you really want confirmation that there were no other calls. You could do this either by counting the callback invocations (with `multiprocessing.Value`) or by accumulating `ctx.triggered` (take the previous `dmc-triggered.children` as `State` and add the new one to it).

PR #657, opened 2025-09-16
update for TipTap V3

### LINKED ISSUE #617: [Feature Request] Rich Text Editor - Increase/Decrease Font Size Button
It would be very helpful to add two buttons to increase and decrease the test size of selected text

AnnMarieW:
Closes #617 

- Updated for TipTap 3 as recommended in Mantine 8.3.0.  Followed [Migration Guide](https://mantine.dev/guides/tiptap-3-migration/)

- Added new features available in TIpTap 3 text style extension:   BackgroundColor, FontFamily, FontSize, LineHeight

- Will add example to the docs for custom FontSize buttons:

```python

from dash import Dash

app = Dash()

component = dmc.RichTextEditor(
    html="<p>Change font size with the controls!</p>",
    toolbar={
        "controlsGroups": [
            ["H1", "H2", "H3", "H4"],
            [
                {
                    "CustomControl": {
                        "aria-label": "Increase font size",
                        "title": "Increase font size",
                        "children": DashIconify(icon="mdi:format-font-size-increase", width=16),
                        "onClick": {"function": "increaseFontSize"},
                    },
                },
                {
                    "CustomControl": {
                        "aria-label": "Decrease font size",
                        "title": "Decrease font size",
                        "children": DashIconify(icon="mdi:format-font-size-decrease", width=16),
                        "onClick": {"function": "decreaseFontSize"},
                    },
                },
            ],
         ],
    },
)


    component
)

if __name__ == "__main__":
    app.run(debug=True)

```

```javascript

var dmcfuncs = window.dashMantineFunctions = window.dashMantineFunctions || {};

function changeFontSize(editor, delta) {
  if (!editor) return;
  const { from, to } = editor.state.selection;
  let size = 16; // default

  editor.state.doc.nodesBetween(from, to, (node) => {
    if (node.isText) {
      const mark = node.marks.find(m => m.type.name === "textStyle");
      if (mark?.attrs.fontSize) {
        size = parseInt(mark.attrs.fontSize, 10);
      }
    }
  });

  const newSize = Math.min(Math.max(size + delta, 8), 72) + "px";
  editor.chain().focus().setFontSize(newSize).run();
}

dmcfuncs.increaseFontSize = ({ editor }) => changeFontSize(editor, 2);
dmcfuncs.decreaseFontSize = ({ editor }) => changeFontSize(editor, -2);
```

dmc-rte-font-size

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Change Log
 
-# [Unreleased]
+# 2.3.0
 
 
-### Added
+### Added  
+
 - New `scrollTo` prop for `ScrollArea` #645 by @AnnMarieW
 - New `ScrollAreaAutoheight` component. #645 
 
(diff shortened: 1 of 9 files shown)


alexcjohnson (approved):
💃 Nice! I presume TipTap v3 has no breaking changes in the html or json data formats it exposes? Those would be the only breaking changes that matter as far as Dash users are concerned.

alexcjohnson:
Unless you expect the next major soon, I think it's fine to include here. This will be generally helpful, and the potential breaking changes are quite deep in custom functionality, with the exception of those small styling changes which I agree seem more like bug fixes. Just make sure to call out in the changelog and docs where these changes might be needed and where to read about them.

AnnMarieW:
OK since there are no known breaking changes in our documented features, it should be OK to include this now.  No plans for a major release any time soon.   I'll update the migration guide and changelog.

Thanks for your guidance on this!

PR #663, opened 2025-10-24
add code highlight to RichTextEditor

AnnMarieW:
This PR adds code highlighting support to `RichTextEditor`, consistent with the [Mantine example](https://mantine.dev/x/tiptap/#code-highlight)

Notes
  - by default, the extension "CodeBlockLowlight" is enabled.
  
  - To keep the bundle size low only a few languages are included ( ts, js, python, css, text)
  
  - When users customize the `extensions` prop, they must explicitly include "CodeBlockLowlight" to keep code highlighting support.

examples:
```python
     # by default all extensions are included
    dmc.RichTextEditor()

    # This has customized extensions and only StarterKit is enabled.  No code highlighting
   dmc.RichTextEditor(
       extensions=["StarterKit"]
    )

    #  This customized extension has code highlighting 
     dmc.RichTextEditor(
       extensions=["StarterKit", "CodeBlockLowlight"]
    )
```

### Example App

```python

from dash import Dash
app = Dash()

component= dmc.RichTextEditor(
    html="",
    extensions=["StarterKit", "CodeBlockLowlight"],
    toolbar={
        "controlsGroups": [
            [
                "Bold",
                "Italic",
                "Underline",
                "Strikethrough",
                "CodeBlock"
            ],
        ],
    },   
)

    component
)

if __name__ == "__main__":
    app.run(debug=True)
```

image

### DIFF
--- a/src/ts/components/extensions/richtexteditor/fragments/RichTextEditor.tsx
+++ b/src/ts/components/extensions/richtexteditor/fragments/RichTextEditor.tsx
@@ -242,7 +242,6 @@ const RichTextEditor = ({
                 // Validate and return other string extensions
                 if (!(ext in extensionMap)) {
                     throw new Error(`Unknown extension: "${ext}"`);
-                    return null;
                 }
                 return extensionMap[ext];
             }
@@ -250,14 +249,12 @@ const RichTextEditor = ({
             // Case 2: Object extension
             if (!ext || typeof ext !== 'object') {
                 throw new Error(`Invalid extension format: ${ext}`);
-                return null;
             }
 
             const keys = Object.keys(ext);
 
             if (keys.length === 0) {
                 throw new Error('Empty extension object');
-                return null;
             }
 
             if (keys.length > 1) {
@@ -270,7 +267,6 @@ const RichTextEditor = ({
 
             if (!(name in extensionMap)) {
                 throw new Error(`Unknown extension: "${name}"`);
-                return null;
             }
 
             // Special handling for CodeBlockLowlight - merge with lowlight


src/ts/components/extensions/richtexteditor/fragments/RichTextEditor.tsx
```diff
@@ -203,14 +220,39 @@ const RichTextEditor = ({
...
+        }
+
+        // Case 2: Object extension (e.g. { CodeBlockLowlight: {} })
+        const name = Object.keys(ext)[0];
+        const opts = ext[name];
+
+        if (name === 'CodeBlockLowlight') {
+            // Merge user options with lowlight
+            return { CodeBlockLowlight: { lowlight, ...opts } };
+        }
```
alexcjohnson: I see this `Object.keys(ext)[0]` pattern below as well, but it seems weird - what if `{}` was passed in? Or what if multiple extensions were passed in as a single object?

Cleaner might be something like:
```js
if ('CodeBlockLowlight' in ext) {
    return {...ext, CodeBlockLowlight: { lowlight, ...ext.CodeBlockLowlight } };
}
```
(or omit the `...ext,` at the beginning if you're confident there will only be one key)

Also below is it acceptable that using an unrecognized name will result in a JS error, from the line `extensionMap[name].configure(ext[name])`? It'll likely be something cryptic like "TypeError: undefined is not an object".

src/ts/components/extensions/richtexteditor/fragments/RichTextEditor.tsx
```diff
@@ -28,13 +28,28 @@ import {
...
     newRenderDashComponent,
     getContextPath,
 } from '../../../../utils/dash3';
 
+const lowlight = createLowlight();
+lowlight.register({ ts });
+lowlight.register('js', js);
+lowlight.register('python', python);
+lowlight.register('css', css);
+lowlight.register('text', plaintext);
```
alexcjohnson: If I'm understanding this right, can't this be written:
```suggestion
lowlight.register({ ts, js, python, css, plaintext });
```
?

Also I wonder if you want to include bash and/or shell, for documenting CLI usage?

src/ts/components/extensions/richtexteditor/fragments/RichTextEditor.tsx
```diff
@@ -203,14 +226,66 @@ const RichTextEditor = ({
...
+                if (ext === 'CodeBlockLowlight') {
+                    return extensionMap.CodeBlockLowlight.configure({
+                        lowlight,
+                    });
+                }
+
+                // Validate and return other string extensions
+                if (!(ext in extensionMap)) {
+                    throw new Error(`Unknown extension: "${ext}"`);
+                    return null;
```
alexcjohnson: You won't get here - after throwing the error, execution stops. Seems fine to me since that would have been the effect before this PR (just remove the unnecessary `return null`), but if you'd prefer this to just make a console error and keep going (dropping the unknown extension by returning `null`) then use `console.error` rather than `throw new Error(...)`

AnnMarieW: I started with `console.warn`, but then changed to throwing an error and forgot to remove the `return null`.  I'll fix that.  I think it's better to use the error because it shows up in the Dash dev tools.  I'm probably not the only one who forgets to  check the console when debugging.

Issue #613, opened 2025-07-08
[Bug] Stepper flickers on update


AnnMarieW:
The Stepper flickers on update in DMC v3 and Dash 3.  See more info in https://github.com/snehilvj/dash-mantine-components/discussions/610,

Thanks for reporting @chgiesse

PR #664, opened 2025-10-24
Fix 613 - remove random react key generation 

chgiesse:
Hey! 
After some lucky coincidence and investigation, I think I found the issue leading to the rendering issues. I just had to change the `key` prop in `newDashRenderComponent` from `Math().random()` to a stable value (`index ?? 0`). 

- Random keys cause React to treat components as completely new instances on every render 
- Stable keys allow React to properly identify and reconcile existing components
- React can efficiently update only changed props instead of destroying and recreating components

you can find details in the official react docs https://react.dev/learn/rendering-lists#rules-of-keys 

I could notice overall rendering performance improvements and these rendering artefacts mentioned in the issue are removed to 99.9%. 

Best regards,
Christian

### DIFF
--- a/src/ts/components/charts/fragments/AreaChart.tsx
+++ b/src/ts/components/charts/fragments/AreaChart.tsx
@@ -3,7 +3,7 @@ import '@mantine/charts/styles.css';
 import React, { useState, useRef } from 'react';
 import { getClickData, isEventValid } from '../../../utils/charts';
 import { getLoadingState } from '../../../utils/dash3';
-import { resolveProp } from '../../../utils/prop-functions';
+import { parseFuncProps } from '../../../utils/prop-functions';
 import { Props } from '../AreaChart';
 
 /** AreaChart */
@@ -19,8 +19,8 @@ const AreaChart = ({
     areaChartProps,
     activeDotProps,
     areaProps,
-    valueFormatter,
-    tooltipProps,
+    data,
+    dataKey,
     ...others
 }: Props) => {
     const [highlightedArea, setHighlightedArea] = useState(null);
@@ -108,6 +108,9 @@ const AreaChart = ({
     return (
         <MantineAreaChart
             data-dash-is-loading={getLoadingState(loading_state) || undefined}
+            {...parseFuncProps('AreaChart', others)}
+            data={data}
+            dataKey={dataKey}
             areaChartProps={newProps}
             series={series}
             activeDotProps={{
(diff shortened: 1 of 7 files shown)


BSd3v:
You'll need to check against the Switch, this was a major issue there.

chgiesse:
I remember, but works, also in a menu. I ran the tests and dmc docs with the index. 

Here are the test results: 
Results (373.22s):
      90 passed
       9 failed
         - tests/dates/test_date_picker.py:73 test_001dp_DatePickerInput
         - tests/dates/test_date_picker.py:136 test_002dp_DatePickerInput
         - tests/dates/test_month_picker.py:65 test_001dp_MonthPickerInput
         - tests/dates/test_month_picker.py:126 test_002dp_MonthPickerInput
         - tests/test_carousel.py:7 test_001ca_carousel
         - tests/test_carousel.py:47 test_002ca_carousel
         - tests/test_new_notification.py:11 test_001na_notification_new_noreopen
         - tests/test_new_notification.py:94 test_003nc_notification_new_clear
         - tests/test_sliders.py:11 test_001sl_slider
       2 skipped

AnnMarieW:
Hi @chgiesse 

Thanks for the PR!  It makes sense that it would be an improvement.
Are you referring to the rendering issue with how the Stepper flickers when updated in a callback? (as described issue  #613)

I tried the Stepper example in the docs both with the current release and with your change, and I don't see much improvement.  It's usually smoother when run locally, rather than in the live docs site, but I still see the flicker.     Can you describe the difference you see and how you measured the performance improvement?

alexcjohnson (approved):
💃 Looks good, thanks for the tweak @chgiesse!

chgiesse:
Hi! I just wanted to checkin on a snippet from the documentation 

-  _"You might be tempted to use an item’s index in the array as its key. In fact, that’s what React will use if you don’t specify a key at all. But the order in which you render items will change over time if an item is inserted, deleted, or if the array gets reordered. Index as a key often leads to subtle and confusing bugs."_ 

That shouldn't be a problem, because the dashRenderer renders the whole list of items also with patch operations - right?  I have multiple callbacks with patch operations on mantine components with the updated version and they work as expected.

AnnMarieW:
@alexcjohnson - thanks for double checking

@chgiesse - thanks again for this PR - it's a nice performance improvement for rendering components-as-props

PR #666, opened 2025-11-02
Rte expose editor

AnnMarieW:
This PR exposes the RichTextEditor editor instance using `dash_mantine_components.getEditor(id) `.  This gives access to all the editor commands in the Tiptap API  https://tiptap.dev/docs/editor/api/commands

Note - the test can be combined with #665 when it's merged.

Here's a cool use-case that can be used in the docs - a character and word counter for the content:

```python
from dash import Dash, Input, Output

app=Dash()
    dmc.RichTextEditor(id="editor", html=""),
    dmc.Button("Get Stats", id="btn-stats", n_clicks=0),
    dmc.Box(id="stats"),
])

app.clientside_callback(
    """
    function(n_clicks) {
        if (n_clicks > 0) {
            const editor = dash_mantine_components.getEditor('editor');
            if (editor) {
                const text = editor.getText();
                const chars = text.length;
                const words = text.split(/\\s+/).filter(Boolean).length;
                return `Characters: ${chars} | Words: ${words}`;
            }
        }
        return dash_clientside.no_update;
    }
    """,
    Output("stats", "children"),
    Input("btn-stats", "n_clicks"),
)


if __name__ == "__main__":
    app.run(debug=True)
```

### DIFF
--- a/tests/test_rich_text_editor.py
+++ b/tests/test_rich_text_editor.py
@@ -278,9 +278,9 @@ def test_004_rich_text_editor_focus_and_readonly(dash_duo):
     updated = updated + " MIDDLE"
     dash_duo.wait_for_text_to_equal(".tiptap", updated)
     
-    # access editor instance in clientsie callback
+    # access editor instance in clientside callback
     dash_duo.find_element('#btn-clientside').click()
-    updated = 'more content' + updated
+    updated = updated + 'more content'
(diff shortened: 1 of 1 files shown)


AnnMarieW:
Thanks @BSd3v for your help getting the updated version working :hugs:

PR #718, opened 2026-04-26
update navlink

### LINKED ISSUE #678: NavLink Active
NavLink active doesnt support:
- Search strings
- URL Quote patterns (spaces escaped to %20 when the href gets updated)

### LINKED ISSUE #717: [Feature Request] Support parent active state based on nested children NavLinks
Thanks for adding the feature for managing active NavLinks without a callback. With the current implementation if a NavLink has children and an href set, when clicking it expands and links to the href. This means its not possible to just have a parent NavLink with active state handling that only opens and closes the children without changing the path. The use case for this is sections of NavLinks where the children are all pages but the parent section should not actually link to a page, but should reflect the active status of its children.

My suggested implementation is to allow the parent of nested NavLinks to inherit its active status from the children. This way if any of the children are active, the parent is active too. Perhaps an option could be there so the active status of the parent is only shown when collapsed. In other words, in the collapsed state the NavLink shows active if any of its children are active, and in the expanded state it does not show active as the actual child state is on-screen.

Alternatively it could be based on the current partial path matching but not actually create the href link when there are children.

AnnMarieW:
Adds two features to `dmc.NavLink` to improve route-based active state without requiring callbacks.

1.  `activeHref`
closes #717

* New prop used only for active state matching.
* Overrides `href` for determining whether a link is active.
* Enables non-navigable parent links (like section headers) to reflect the active state of their children.

Example:

```python
dmc.NavLink(
    label="Reports",
    active="partial",
    activeHref="/reports",
    children=[...],
)
```

2. `exact-with-search`
closes #678

* New option for the `active` prop.
* Marks a link as active only when both the pathname and query string match.
* Query parameters are compared after decoding and are order-insensitive.

Example:

```python
dmc.NavLink(
    label="Sales",
    href="/reports?type=Sales",
    active="exact-with-search",
)
```

**Additional notes**

* `active=True` and `active=False` still override all matching behavior.
* Existing `exact` and `partial` modes are fixed so they ignore query parameters for match
* Click behavior for links with `children` is updated so they can act as non-navigable toggles when `href` is not provided.

Here is a complete minimal example:

```python

import dash
from dash import Dash, html

app = Dash(use_pages=True, pages_folder="")


dash.register_page("home", path="/", layout=html.Div("Home"))

def reports_layout(type=None, **kwargs):
    return html.Div(f"{type} Report ")

dash.register_page("reports", path="/reports", layout=reports_layout)
dash.register_page("settings", path="/settings", layout=html.Div("Settings"))


component = dmc.Box([
    dmc.NavLink(label="Home", href="/", active="exact"),

    dmc.NavLink(
        label="Reports",
        # non-navigable parent
        active="partial",
        activeHref="/reports",
        childrenOffset=28,
        persistence=True,
        children=[
            dmc.NavLink(
                label="Sales",
                href="/reports?type=Sales",
                active="exact-with-search",
            ),
            dmc.NavLink(
                label="Inventory",
                href="/reports?type=Inventory",
                active="exact-with-search",
            ),
        ],
    ),

    dmc.NavLink(label="Settings", href="/settings", active="exact"),

    dmc.Divider(mb="lg"),
    dash.page_container
])

if __name__ == "__main__":
    app.run(debug=True)



```

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Added  `activeHref` for custom active-state matching independent of navigation, and introduce `exact-with-search` to match both pathname and query parameters. #718 by @AnnMarieW
 - Added `buttonProps` to `Button`, `ActionIcon`, `UnstyledButton`, `CopyButton` components to allow passing props directly to the underlying button element. #716 by @AnnMarieW
 
 # 2.6.1
--- a/src/ts/components/core/NavLink.tsx
+++ b/src/ts/components/core/NavLink.tsx
@@ -26,14 +26,22 @@ interface Props
     rightSection?: React.ReactNode;
     /**
      * Controls whether the link is styled as active (default: `false`).
-     * - `exact`: Active if `pathname` matches `href` exactly.
-     * - `partial`: Active if `pathname` starts with `href` (for subpages).
+     *
+     * - `False`: never active, overrides all matching behavior.
+     * - `True`: always active, overrides all matching behavior.
+     * - `exact`: active when the current `pathname` matches `href` pathname exactly.
+     * - `partial`: active when the current `pathname` starts with `href` pathname.
+     * - `exact-with-search`: active when both `pathname` and query parameters
+     *   match `href`. Query parameters are compared after decoding and are
+     *   order-insensitive.
      */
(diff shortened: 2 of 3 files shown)


AnnMarieW:
closed in favor of #719

Issue #723, opened 2026-05-27
Enable functions in `valueLabelProps` in `BarChart`


AnnMarieW:
To make it possible to create custom bar value labels (similar to the custom tooltips), it's necessary to enable functions in the `valueLabelProp`.   Then a custom component can be passed to the underlying Recharts LableList component.

Just need to add `valueLabelProp` here:  https://github.com/snehilvj/dash-mantine-components/blob/master/src/ts/utils/prop-functions.ts#L66

almostBurtMacklin:
Created pull request for fix #724

AnnMarieW:
Closed by #724

PR #724, opened 2026-05-27
Add 'valueLabelProp' to BarChart properties

almostBurtMacklin:
Allows for custom component for the label prop - fix for issue: #723

### DIFF
--- a/src/ts/utils/prop-functions.ts
+++ b/src/ts/utils/prop-functions.ts
@@ -64,7 +64,7 @@ const funcPropsMap = {
         'gridProps',
         'rightYAxisProps',
         'tooltipProps',
-        'valueLabelProp',  
+        'valueLabelProps',  
         'getBarColor',
     ],
     BubbleChart: [
--- a/src/ts/utils/prop-functions.ts
+++ b/src/ts/utils/prop-functions.ts
@@ -64,6 +64,7 @@ const funcPropsMap = {
         'gridProps',
         'rightYAxisProps',
         'tooltipProps',
+        'valueLabelProp',  
         'getBarColor',
     ],
     BubbleChart: [
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Change Log
 
+# Unreleased
+
+### Added
+
+- Enable functions in valueLabelProps in BarChart  #724 by @almostBurtMacklin 
+
 # 2.7.0
 
 ### Added


src/ts/utils/prop-functions.ts
```diff
@@ -64,6 +64,7 @@ const funcPropsMap = {
         'gridProps',
         'rightYAxisProps',
         'tooltipProps',
+        'valueLabelProp',
```
AnnMarieW:
```suggestion
        'valueLabelProps',  
```

AnnMarieW:
Hi @almostBurtMacklin 

Thanks for the PR!

Can you add a complete minimal sample app to show it works?  


It should follow the same pattern as the custom tooltip.  https://www.dash-mantine-components.com/components/barchart#custom-tooltip

Might look something like this for the bar chart:
```python
dmc.BarChart(
    h=300,
    dataKey="price_code",
    data=record,
    type="stacked",
    withBarValueLabel=True,
    valueLabelProps={
        "content": {"function": "barValueLabel"}
    },
    series=[
        {"name": "HELD BACK", "color": "red.6"},
        {"name": "TM", "color": "blue.6"},
    ],
)
```

in /assets .js

```js

var dmcfuncs = window.dashMantineFunctions =
  window.dashMantineFunctions || {};

dmcfuncs.barValueLabel = (props) => {
  const { x, y, width, height, value } = props;

  return React.createElement(
    "text",
    {
      x: x + width / 2,
      y: y + height / 2,
      fill: "white",
      textAnchor: "middle",
      dominantBaseline: "middle",
      fontSize: 12,
    },
    value
  );
};
```

almostBurtMacklin:
Hi @AnnMarieW 

Sorry for the typo on the pull request... anyways I corrected it and installed the new version locally to test.

After investigating further, I believe the behavior I am seeing is due to a limitation in Recharts rather than the Dash wrapper implementation.

In stacked BarChart mode, the LabelList component does not expose per-series (payload) values. Instead, it only receives the computed stacked value (value), which represents the cumulative total at that point in the stack (e.g. 100 -> 300).

This is confirmed by logging the runtime props, which only include geometry and the aggregated value, but not the individual dataKey value for the specific stack segment.

In contrast, tooltips behave differently because they operate in a data-access context and receive the full payload for the hovered data, including all series values for the row. This allows tooltip customizations to access raw values per series, which is not available in LabelList.

Because of this difference in execution context:

Tooltip - data-layer component (has full payload / raw values)
LabelList - render-layer component (only receives computed geometry + aggregated stack value)

As a result, even with function-enabled valueLabelProps, there is no reliable way to retrieve the individual bar segment value from LabelList in stacked mode at the chart level.

The current implementation works correctly for passing custom render functions, but in stacked mode the underlying charting library does not provide the necessary per-series value context to render true “segment-level” labels.

A possible improvement (outside the scope of this PR) would be to support applying LabelList per <Bar dataKey> instead of at the stacked chart level, which would restore access to correct per-series values. That could look like something below:

`series=[
      {
          "name": "sales",
          "color": "blue",
          "valueLabelProps": {
              "function": "barValueLabel"
          }
      },
      {
          "name": "revenue",
          "color": "red",
          "valueLabelProps": {
              "function": "barValueLabel"
          }
      },
      ],`

TL;DR: because Recharts does not expose payload in dmcfuncs.barValueLabel (like it does in dmcfuncs.chartTooltip) I don't think we can display the value for each stack/slice of stack bar charts with a .js function

AnnMarieW:
Oh, I see what you mean.
We can still keep your PR though, since there are other uses for functions within `valueLabelProps`


For your use-case would a Plotly stacked bar chart work?  There is some info on styling with a Mantine theme in the docs here:  https://www.dash-mantine-components.com/components/figure-templates

almostBurtMacklin:
I will probably just stick with dmc.BarChart without the labels. The tooltip customization is more valuable than labels at this point. I appreciate the help on this. I will keep an eye on Recharts and see if they make any changes to their props.

almostBurtMacklin:
Here is a MWE for the new bar chart value customization: https://github.com/almostBurtMacklin/BarChartMWE

AnnMarieW:
OK - thanks!

If you would like to add an entry in CHANGELOG.md entry, we can get this merged.   At the top of the file add

```
# Unreleased

### Added

- Enable functions in valueLabelProps in BarChart  #724 by @almostBurtMacklin 

```

AnnMarieW:
Thanks for the PR!   
Sorry it didn't work for your use-case, but it's still a nice addition.

